PR #3, opened 2018-04-25
support coprocessor request.

coocood:
It's runnable now.
Most of the code is copied from `mocktikv`.
Need to rename `vendor` under `tidb` to `_vendor` to build.


ngaut:
Cool

PR #21, opened 2018-05-25
serialize write operation

coocood:


ngaut:
How about 3M rows?

tikv/region.go
```diff
@@ -109,6 +119,32 @@ func (ri *regionInfo) marshal() []byte {
 	return data
 }
 
+func (ri *regionCtx) lock(hashVals []uint64) (bool, *sync.WaitGroup) {
```
ngaut: s/lock/aquireLocks

tikv/region.go
```diff
@@ -109,6 +119,32 @@ func (ri *regionInfo) marshal() []byte {
...
+	}
+	wg := new(sync.WaitGroup)
+	wg.Add(1)
+	for _, hashVal := range hashVals {
+		ri.memLocks[hashVal] = wg
+	}
+	return true, nil
+}
+
+func (ri *regionCtx) unlock(hashVals []uint64) {
```
ngaut: s/unlock/releaseLocks/

coocood:
@ngaut PTAL

PR #22, opened 2018-05-28
refactor for efficient Resolve request.

coocood:


tikv/region.go
```diff
@@ -145,6 +154,45 @@ func (ri *regionCtx) releaseLocks(hashVals []uint64) {
...
+				copy(newKeys, keys[:i])
+				copy(newKeys[i:], keys[i+1:])
+				ri.txnKeysMap[startTS] = newKeys
+			}
+		}
+	}
+	ri.txnKeysMu.Unlock()
+}
+
+func (ri *regionCtx) geTxnKeys(startTS uint64) [][]byte {
```
ngaut: s/geTxnKeys/getTxnKeys/

tikv/region.go
```diff
@@ -145,6 +154,45 @@ func (ri *regionCtx) releaseLocks(hashVals []uint64) {
...
+	ri.txnKeysMu.RLock()
+	keys := ri.txnKeysMap[startTS]
+	ri.txnKeysMu.RUnlock()
```
ngaut: There is nothing left behind lock operation, use defer instead.

PR #23, opened 2018-05-29
multiple improvements

coocood:
1. Avoid exceeds max transaction size by split entries into batches.
2. fix lock bug in prewrite and commit.
3. fix table scan and index scan bugs.
4. fix reverse scan bug
5. use getter to avoid creating iterator in BatchGet


coocood:
@ngaut PTAL

PR #31, opened 2018-06-10
introduce lockstore

coocood:
A in-memory lock-free data-structure to store lock.
- allows single writer and multiple readers without lock.
- memory can be reused after all the entries in a block are deleted.

### DIFF
--- a/lockstore/lockstore_test.go
+++ b/lockstore/lockstore_test.go
@@ -1,0 +1,203 @@
+package lockstore
+
+import (
+	"bytes"
+	"fmt"
+	"math/rand"
+	"testing"
+	"time"
+
+	"github.com/stretchr/testify/require"
+)
+
+func TestLockStore(t *testing.T) {
+	prefix := "ls"
+	n := 30000
+	ls := NewLockStore(1 << 10)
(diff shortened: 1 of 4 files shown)


lockstore/lockstore_test.go
```diff
@@ -0,0 +1,203 @@
...
+	b.ResetTimer()
+	for i := 0; i < b.N; i++ {
+		n := r.Intn(10000)
+		ls.Delete(keys[n])
+		ls.Insert(keys[n], keys[n])
+		ls.Get(keys[n], buf)
+	}
+}
+
+func BenchmarkLockStoreIterate(b *testing.B) {
```
ngaut: How does it compare to original skip list in badger?

coocood:
@ngaut 
single thread run 1 million keys put and 1 million keys get.

skl put 637ns
skl get 570ns
lockstore insert 605ns
lockstore get 837ns

coocood:
@ngaut after manual inline `getNext` function. the performance improved a lot.

skl put 618ns
skl get 598ns
lockstore insert 594ns
lockstore get 666ns

PR #35, opened 2018-06-14
avoid holding txnKeysMu for long time

coocood:
We don't need to inherent txn keys from parent, we can add them when client encounter it.


tikv/region.go
```diff
@@ -455,29 +455,7 @@ func (rm *RegionManager) getRegionFromCtx(ctx *kvrpcpb.Context) (*regionCtx, *er
...
-						}
-					}
-				}
-				if len(newKeys) > 0 {
-					ri.txnKeysMap[startTS] = newKeys
-				}
-			}
-			ri.txnKeysMu.Unlock()
-		}
+		atomic.StorePointer(&ptr, nil)
```
ngaut: Could add a todo here? We are moving to next stage, we still need to handle lock split.

coocood:
@ngaut PTAL

PR #39, opened 2018-06-19
Make code more clear

ngaut:


tikv/mvcc.go
```diff
@@ -58,20 +58,27 @@ func NewMVCCStore(db *badger.DB, dataDir string) *MVCCStore {
...
 	go store.writeLockWorker.run()
-	rbGCWorker := rollbackGCWorker{store: store}
-	go rbGCWorker.run()
+	go func() {
+		rbGCWorker := rollbackGCWorker{store: store}
+		rbGCWorker.run()
+	}()
+
+	// mark worker count
+	store.wg.Add(3)
```
coocood: It's safer to add WaitGroup count before the goroutines run.

ngaut: Why was that? we are still inside the current initialize function.

coocood: In this case, it is safe, but in general coding style, we may run into wg count decrease to a negative value in some cases.

tikv/mvcc.go
```diff
@@ -58,20 +58,27 @@ func NewMVCCStore(db *badger.DB, dataDir string) *MVCCStore {
...
 }
 
 func (store *MVCCStore) Close() error {
 	close(store.closeCh)
 	store.wg.Wait()
+
 	err := store.dumpLockStore()
 	if err != nil {
-		log.Error(err)
+		log.Fatal(err)
```
coocood: We need to close the DB later.

ngaut: dumpLockStore should never fail.

tikv/mvcc.go
ngaut: We should never use WaitGroup until we initialize it successful. I prefer to keep this clear.

PR #45, opened 2018-06-20
check `badger.ErrKeyNotFound` error returned by `txn.Get`.

coocood:


tikv/mvcc.go
```diff
@@ -265,9 +265,12 @@ func (store *MVCCStore) Commit(req *requestCtx, keys [][]byte, startTS, commitTS
 func (store *MVCCStore) handleLockNotFound(reqCtx *requestCtx, key []byte, startTS, commitTS uint64) error {
 	txn := reqCtx.getDBReader().txn
 	item, err := txn.Get(key)
-	if err != nil {
+	if err != nil && err != badger.ErrKeyNotFound {
 		return errors.Trace(err)
 	}
+	if item == nil {
```
ngaut: Check if err if badger.ErrKeyNotFound  should be more clear.

tikv/mvcc.go
```diff
@@ -339,7 +342,7 @@ func (store *MVCCStore) rollbackKey(req *requestCtx, batch *writeBatch, key []by
 	}
 	reader := req.getDBReader()
 	item, err := reader.txn.Get(key)
-	if err != nil {
+	if err != nil && err != badger.ErrKeyNotFound {
```
ngaut: What happend if err is badger.ErrKeyNotFound?

coocood: Then `item` would be `nil`

tikv/mvcc.go
coocood: I think `badger.ErrKeyNotFound` is redundant, check if the returned object is nil is a more commonly used pattern.

PR #61, opened 2018-07-11
make it compile with latest tidb

ngaut:


tikv/cop_handler.go
```diff
@@ -67,8 +69,8 @@ func (svr *Server) handleCopDAGRequest(reqCtx *requestCtx, req *coprocessor.Requ
 		chunks = appendRow(chunks, data, rowCnt)
 		rowCnt++
 	}
-	warnings := dagCtx.evalCtx.sc.GetWarnings()
-	return buildResp(chunks, e.Counts(), err, warnings, time.Since(startTime))
+	//warnings := dagCtx.evalCtx.sc.GetWarnings()
+	return buildResp(chunks, e.Counts(), err, nil, time.Since(startTime))
```
coocood: We can change the `buildResp` `warnings` arg type as
```
func buildResp(chunks []tipb.Chunk, counts []int64, err error, warnings []stmtctx.SQLWarn, dur time.Duration) *coprocessor.Response {
```

coocood: and change append warnings as
```
selResp.Warnings = append(selResp.Warnings, toPBError(warnings[i].Err))
```

ngaut: Thank you for your comments. Addressed.

tikv/cop_handler.go
```diff
@@ -40,6 +40,7 @@ type dagContext struct {
 func (svr *Server) handleCopDAGRequest(reqCtx *requestCtx, req *coprocessor.Request) *coprocessor.Response {
 	startTime := time.Now()
 	resp := &coprocessor.Response{}
+	// TODO: handle warning
```
coocood: The `TODO` can be removed?

PR #71, opened 2018-07-25
optimize count(*) and DecodeHandle.

coocood:
1. Do not decode the row if only handle column is needed.

2, Implement a more efficient decodeRowKeyHandle.

### DIFF
--- a/tikv/cop_handler.go
+++ b/tikv/cop_handler.go
@@ -216,6 +216,11 @@ func (svr *Server) buildTableScan(ctx *dagContext, executor *tipb.Executor) (*ta
 		buf:       make([][]byte, len(columns)),
 		loc:       ctx.evalCtx.sc.TimeZone,
 	}
+	if len(columns) == 1 {
+		if columns[0].PkHandle || columns[0].ColumnId == model.ExtraHandleID {
+			e.handleOnly = true
+		}
+	}
 	if ctx.dagReq.CollectRangeCounts != nil && *ctx.dagReq.CollectRangeCounts {
 		e.counts = make([]int64, len(ranges))
 	}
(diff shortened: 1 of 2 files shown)


tikv/executor.go
```diff
@@ -298,6 +303,40 @@ func (e *tableScanExec) fillChunkFromRange(chk *chunk.Chunk, ran kv.KeyRange) (e
 	return
 }
 
+const (
```
ngaut: Move the changes to tidb.

coocood: I'll do it later.

ngaut: Let's keep this PR clean.

coocood:
@ngaut PTAL

tikv/cop_handler.go
```diff
@@ -216,6 +216,11 @@ func (svr *Server) buildTableScan(ctx *dagContext, executor *tipb.Executor) (*ta
 		buf:       make([][]byte, len(columns)),
 		loc:       ctx.evalCtx.sc.TimeZone,
 	}
+	if len(columns) == 1 {
+		if columns[0].PkHandle || columns[0].ColumnId == model.ExtraHandleID {
```
ngaut: Do we have a function to check if a column is PK handle in tidb, we should reuse it if we do have.

ngaut:
Rest LGTM.

tikv/cop_handler.go
coocood: TiDB didn't extract the logic to a function to check the column.

coocood: Actually, the second check can be removed.

coocood:
@ngaut PTAL

PR #76, opened 2018-07-28
implement closure executor

coocood:


tikv/closure_exec.go
```diff
@@ -0,0 +1,162 @@
...
+	"github.com/juju/errors"
+	"github.com/pingcap/tipb/go-tipb"
+	"github.com/pingcap/tidb/util/codec"
+)
+
+// tryBuildClosureExecutor tries to build a closureExecutor for the DAGRequest.
+// currently, only 'count(*)' is supported, but we can support all kinds of requests in the future.
+func (svr *Server) tryBuildClosureExecutor(dagCtx *dagContext, dagReq *tipb.DAGRequest) *closureExecutor {
+	executors := dagReq.Executors
+	if len(executors) != 2 {
```
ngaut: Please add comments why compares to 2?

tikv/closure_exec.go
```diff
@@ -0,0 +1,162 @@
...
+	var unique bool
+	switch executors[0].Tp {
+	case tipb.ExecType_TypeTableScan:
+		tblScan := executors[0].TblScan
+		cols = tblScan.Columns
+		unique = true
+	case tipb.ExecType_TypeIndexScan:
+		idxScan := executors[0].IdxScan
+		cols = idxScan.Columns
+		unique = idxScan.GetUnique()
```
ngaut: Should we return nil for other Type in default branch?

tikv/closure_exec.go
coocood: This will be removed after we support simple IndexScan and TableScan.

tikv/closure_exec.go
coocood: The first executor is always IndexScan or TableScan.

tikv/closure_exec.go
```diff
@@ -0,0 +1,162 @@
...
+			if err != nil {
+				return nil, errors.Trace(err)
+			}
+		}
+	}
+	err = e.finishFunc()
+	return e.oldChunks, err
+}
+
+func (e *closureExecutor) canUsePointGet(ran kv.KeyRange) bool {
```
ngaut: s/ran/range/

coocood: It's a keyword in Go, doesn't compile.

tikv/closure_exec.go
```diff
@@ -0,0 +1,162 @@
...
+	if !e.unique || len(ran.StartKey) != len(ran.EndKey) {
+		return false
+	}
+	lastIdx := len(ran.StartKey) - 1
+	if int(ran.StartKey[lastIdx]) + 1 != int(ran.EndKey[lastIdx]) {
```
ngaut: The block is not hot path, please use bytes.Equal to compare them.

tikv/closure_exec.go
ngaut: So please add comments for assumption

tikv/closure_exec.go
ngaut: How about simply use r ?

PR #77, opened 2018-07-29
handle end key checking implicitly and remove the checking out of hot…

ngaut:
… path


coocood:
Can we remove all the checks and assume end key never be empty?

coocood:
Because we do the initial split, the region with the empty end key is never used.

ngaut:
I prefer to keep a panic, so if there is a error, we can catch it. I will change all of them to panic.

PR #79, opened 2018-08-13
support the new row format

coocood:
- The old row is converted to the new format in `Prewrite`.
- The new row is converted to the old format if closure executor is not supported for Coprocessor request.
- The new row is converted to the old format in `Scan`, `Get` and `BatchGet`.


coocood:
mysql_test is passed

PR #81, opened 2018-08-16
optimize and refactor encoding and add benchmark

coocood:
Benchmark result for 3 column row (int, string, float):

Encode:
10000000	       114 ns/op	       0 B/op	       0 allocs/op

EncodeFromOldRow:
10000000	       257 ns/op	       0 B/op	       0 allocs/op

Decode:
20000000	        65.9 ns/op	       0 B/op	       0 allocs/op

The `EncodeFromOldRow` is currently used, it first decodes from an old row, then encodes to a new row. When we encode row in TiDB, `Encode` will be used.

### DIFF
--- a/rowcodec/common.go
+++ b/rowcodec/common.go
@@ -25,8 +25,8 @@ const (
 	VarintFlag       byte = 8
 )
 
-// XRow is the struct type used to access the new row format.
-type XRow struct {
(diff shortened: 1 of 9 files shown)


coocood:
@ngaut PTAL

PR #83, opened 2018-09-04
run db.Update in a new goroutine

coocood:
Reduce the blocking time.


coocood:
@ngaut PTAL

ngaut:
Any performance comparison?

PR #87, opened 2018-10-17
correct recordPrefixIdx

bobotu:
rowkey format is `t{8 bytes id}_r{8 bytes id}`.

`IsRowKey` used by `Prewrite`, the incorrect result cause it never convert value to new row format.

### DIFF
--- a/rowcodec/encoder.go
+++ b/rowcodec/encoder.go
@@ -300,7 +300,7 @@ var defaultStmtCtx = &stmtctx.StatementContext{
 
 const (
 	rowKeyLen       = 19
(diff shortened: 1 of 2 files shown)


coocood:
This is introduced in this commit https://github.com/ngaut/unistore/commit/b12001038b041a6c4cf59960571bf966185286fe
Maybe we should revert it to make it work with tidb-server in master version.

coocood:
In that commit, we tried to changed the key format and use multiple DB to improve write performance and it only works with a specific tidb branch.

PR #88, opened 2018-10-18
add a flag to control shard-key feature

bobotu:
Use a flag to control db sharding. Disable it when work with master tidb-server.

### DIFF
--- a/rowcodec/common.go
+++ b/rowcodec/common.go
@@ -13,7 +13,6 @@ import (
 // CodecVer is the constant number that represent the new row format.
 const CodecVer = 128
 
-var EnableShardKeySupport = false
 var invalidCodecVer = errors.New("invalid codec version")
 
 // First byte in the encoded value which specifies the encoding type.
--- a/rowcodec/common.go
+++ b/rowcodec/common.go
@@ -13,6 +13,7 @@ import (
 // CodecVer is the constant number that represent the new row format.
 const CodecVer = 128
 
+var EnableShardKeySupport = false
 var invalidCodecVer = errors.New("invalid codec version")
 
 // First byte in the encoded value which specifies the encoding type.
--- a/rowcodec/encoder.go
+++ b/rowcodec/encoder.go
@@ -299,8 +299,11 @@ var defaultStmtCtx = &stmtctx.StatementContext{
 }
 
 const (
+	// Length of rowkey.
 	rowKeyLen             = 19
+	// Index of record flag 'r' in rowkey used by master tidb-server.
 	recordPrefixIdx       = 10
+	// Index of record flag 'r' in rowkey whit shard byte.
 	sharedRecordPrefixIdx = 1
 )
 
--- a/rowcodec/encoder.go
+++ b/rowcodec/encoder.go
@@ -302,13 +302,14 @@ const (
 	// Length of rowkey.
 	rowKeyLen = 19
 	// Index of record flag 'r' in rowkey used by master tidb-server.
+	// The rowkey format is t{8 bytes id}_r{8 bytes handle}
 	recordPrefixIdx = 10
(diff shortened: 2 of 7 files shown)


rowcodec/encoder.go
```diff
@@ -299,11 +299,15 @@ var defaultStmtCtx = &stmtctx.StatementContext{
 }
 
 const (
-	rowKeyLen       = 19
-	recordPrefixIdx = 1
+	rowKeyLen             = 19
+	recordPrefixIdx       = 10
```
ngaut: Please add comments for these constant numbers.

rowcodec/common.go
```diff
@@ -13,6 +13,7 @@ import (
 // CodecVer is the constant number that represent the new row format.
 const CodecVer = 128
 
+var EnableShardKeySupport = false
```
coocood: How about remove this variable and make two isRowKey functions, I think that would be cleanner, maintains the state in a single place.

bobotu:
* move `enableShardKey` and `EnableShardKeySupport` to `server.go`
* split `IsRowKeyWithShardByte` from `IsRowKey`
* add `isRowKey` in `util.go`, which use correct function in `rowcodec` package based on `enableShardKey`

@coocood PTAL

tikv/region.go
```diff
@@ -157,13 +157,16 @@ func (ri *regionCtx) releaseLatches(hashVals []uint64) {
 }
 
 func (ri *regionCtx) getDBIdx() int {
+	if !enableShardKey {
```
ngaut: s/enableShardKey/EnableShardKeySupport/

tikv/region.go
```diff
@@ -300,18 +303,23 @@ func (rm *RegionManager) initialSplit(root *metapb.Region) {
...
-		{'t', 'i', 64},
-		{'t', 'i', 128},
-		{'t', 'i', 192},
-		{'t', 'r', 0},
-		{'t', 'r', 64},
-		{'t', 'r', 128},
-		{'t', 'r', 192},
-		{'u'},
+	var preSplitStartKeys [][]byte
+	if enableShardKey {
```
ngaut: Ditto.

tikv/server.go
```diff
@@ -16,6 +16,12 @@ import (
 	"golang.org/x/net/context"
 )
 
+var enableShardKey = false
+
+func EnableShardKeySupport() {
```
ngaut: Could you add a function IsShardingEnabled()? And call it instead of using enableShardKey directly.

rowcodec/encoder.go
```diff
@@ -299,10 +299,18 @@ var defaultStmtCtx = &stmtctx.StatementContext{
 }
 
 const (
-	rowKeyLen       = 19
-	recordPrefixIdx = 1
+	// Length of rowkey.
+	rowKeyLen = 19
+	// Index of record flag 'r' in rowkey used by master tidb-server.
```
ngaut: Could you explain why the length of recordPrefixIdx is 10?

tikv/util.go
```diff
@@ -42,3 +43,10 @@ func keysToHashVals(keys ...[]byte) []uint64 {
 func safeCopy(b []byte) []byte {
 	return append([]byte{}, b...)
 }
+
+func isRowKey(key []byte) bool {
+	if enableShardKey {
```
ngaut: Ditto.

rowcodec/encoder.go
bobotu: The rowkey format in master tidb is `t{8 bytes id}_r{8 bytes handle}`. The index of `r` is 10.

bobotu:
@ngaut PTAL

ngaut:
LGTM

PR #91, opened 2018-11-17
use multiple workers

ngaut:


coocood:
Any benchmark result?

PR #92, opened 2018-11-28
use new TxnGroup API to build a write pipeline

bobotu:
- [ ] Update `Gopkg.lock` after badger merged corresponding PR


bobotu:
Want to replace #91 and #83

PR #96, opened 2018-12-21
move startTS and commitTS from mvccValue to UserMeta.

coocood:
So we don't need to read the value from vlog to get the startTS and commitTS.

Prepare for GC.

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -131,7 +131,7 @@ func (store *MVCCStore) Prewrite(reqCtx *requestCtx, mutations []*kvrpcpb.Mutati
 	var buf []byte
 	var enc rowcodec.Encoder
 	for i, m := range mutations {
-		oldVal, err := store.checkPrewriteInDB(reqCtx, txn, m, startTS)
+		oldMeta, oldVal, err := store.checkPrewriteInDB(reqCtx, txn, m, startTS)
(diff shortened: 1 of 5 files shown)


coocood:
@bobotu PTAL

tikv/mvcc.go
```diff
@@ -131,7 +131,7 @@ func (store *MVCCStore) Prewrite(reqCtx *requestCtx, mutations []*kvrpcpb.Mutati
 	var buf []byte
 	var enc rowcodec.Encoder
 	for i, m := range mutations {
-		oldVal, err := store.checkPrewriteInDB(reqCtx, txn, m, startTS)
+		oldMeta, oldVal, err := store.checkPrewriteInDB(reqCtx, txn, m, startTS)
```
bobotu: Just a little curious, can we delay the vlog read to `Commit`, let `checkPrewriteInDB` just return `UserMeta`. So we can reduce some useless vlog read when `anyError == true`, and we don't need to store the old value in `LockStore`.

coocood: Then if there is no error, there would be one more LSM read.

bobotu: Got.

coocood:
@ngaut PTAL

ngaut:
LGTM

PR #98, opened 2018-12-25
implement GCCompactionFilter

coocood:


coocood:
Not fully tested yet, will do it later.

coocood:
I ran a manual test.

1. initialize a new cluster, set the GC parameter to 1 minute.

```
update tidb set variable_value='1m' where variable_name='tikv_gc_run_interval' or variable_name = 'tikv_gc_life_time';
```

2. Insert 3 million sysbench rows, then there are 11 SST files in DB1.


3. Run sysbench oltp_write for 20 minutes.

With compactionFilter enabled, the number of SST files is 22.
Without compactionFilter, the number of SST files is 45.

coocood:
@ngaut PTAL

tikv/mvcc.go
```diff
@@ -618,6 +640,64 @@ func (store *MVCCStore) deleteKeysInBatch(reqCtx *requestCtx, keys [][]byte, bat
...
+		if err != nil {
+			return err
+		}
+		atomic.StoreUint64(&store.safePoint.value, safePoint)
+	}
 	return nil
 }
+
+type SafePoint struct {
+	value uint64
```
ngaut: s/value/timestamp/

tikv/mvcc.go
```diff
@@ -63,6 +67,24 @@ func NewMVCCStore(dbs []*badger.DB, dataDir string) *MVCCStore {
 	if err != nil {
 		log.Fatal(err)
 	}
+	// Load safe point.
+	err = dbs[0].View(func(txn *badger.Txn) error {
```
ngaut: extract to a function

tikv/mvcc.go
```diff
@@ -618,6 +640,64 @@ func (store *MVCCStore) deleteKeysInBatch(reqCtx *requestCtx, keys [][]byte, bat
 }
 
 func (store *MVCCStore) GC(reqCtx *requestCtx, safePoint uint64) error {
-	// TODO: implement GC in badger.
+	// We use the gcLock to make sure safePoint can only increase.
+	store.gcLock.Lock()
```
ngaut: This is the only place that use gcLock? Why GC is called concurrently? And is it really necessary?

tikv/region.go
```diff
@@ -25,6 +25,7 @@ var (
 	InternalKeyPrefix        = []byte(`i`)
 	InternalRegionMetaPrefix = append(InternalKeyPrefix, "region"...)
 	InternalStoreMetaKey     = append(InternalKeyPrefix, "store"...)
+	InternalSafePointKey     = append(InternalKeyPrefix, "sp"...)
```
ngaut: s/sp/saftpoint/

tikv/mvcc.go
coocood: I'm not 100% sure there is no concurrently call from TiDB, just in case.

tikv/mvcc.go
```diff
@@ -618,6 +640,64 @@ func (store *MVCCStore) deleteKeysInBatch(reqCtx *requestCtx, keys [][]byte, bat
...
+// For old keys, if the nextCommitTS is before the safePoint, it means we already have one version before the safePoint,
+// we can safely drop this entry.
+//
+// For latest keys, only Delete entry should be GCed, if the commitTS of the Delete entry is older than safePoint,
+// we can remove the Delete entry. But we need to convert it to a tombstone instead of drop.
+// Because there maybe multiple badger level old entries under the same key, dropping it results in old
+// entries may appear again.
+func (f *GCCompactionFilter) Filter(key, value, userMeta []byte) badger.Decision {
+	switch key[0] {
+	case 'm', 't':
```
ngaut: could you define these magic letters as readable flags?

tikv/mvcc.go
```diff
@@ -618,6 +640,64 @@ func (store *MVCCStore) deleteKeysInBatch(reqCtx *requestCtx, keys [][]byte, bat
...
+		// For latest version, we can only remove `delete` key, which has value len 0.
+		if dbUserMeta(userMeta).CommitTS() < f.safePoint && len(value) == 0 {
+			return badger.DecisionDelete
+		}
+	case 'n', 'u':
```
ngaut: Ditto.

tikv/mvcc.go
```diff
@@ -677,6 +677,13 @@ type GCCompactionFilter struct {
 	safePoint uint64
 }
 
+const (
+	metaPrefix     byte = 'm'
+	metaOldPrefix  byte = 'n'
```
ngaut: please add comments for metaOldPrefix.

tikv/mvcc.go
```diff
@@ -677,6 +677,13 @@ type GCCompactionFilter struct {
...
+	tablePrefix    byte = 't'
+	tableOldPrefix byte = 'u'
```
ngaut: please add comments for tableOldPrefix.

Issue #107, opened 2019-01-12
Support SST ingestion in badger


coocood:
This feature is required for port Snapshot related code in TiKV

coocood:
@bobotu

bobotu:
Will complete this in three steps:
1. Add a RocksDB compatible `SST file writer/reader`.
1. Introduce a new meta field `FileGlobalTS` which represent the TS of keys in the ingested file.
   - Update SST file format.
   - Update `blockIterator` to use the global ts when reading a key from ingested files.
1. Add `IngestExternalFiles` function which converts files into badger's internal SST format and put them to the best level possible in the LSM-tree.

coocood:
We need to add the license at the head of each file in this package.

PR #108, opened 2019-01-21
port SstFileWriter from RocksDB

bobotu:
I have checked some sst files, which generated by this SstFileWriter with different table size and compression algorithm, using RocksDB's sst check tools and `IngestExternalFiles`.   

This writer simplified many options in RocksDB, which doesn't use by TiKV. We can add them back if need.

I can port some unit tests later.

The first step of #107 😁


coocood:
LGTM

PR #109, opened 2019-01-24
add sst file iterator

bobotu:
refer to #107


rocksdb/sst_file_iterator.go
```diff
@@ -0,0 +1,196 @@
...
+	indexBlockIter *blockIterator
+	dataBlockIter  *blockIterator
+	dataBuf        []byte
+	invalid        bool
+	err            error
+	checksumType   ChecksumType
+}
+
+func NewSstFileIterator(f *os.File) *SstFileIterator {
+	return &SstFileIterator{
```
coocood: I think it's better to load the footer in this function.
`SeekToFirst` may be called multiple times.

rocksdb/sst_file_iterator_test.go
```diff
@@ -0,0 +1,116 @@
...
+		key := it.Key()
+		value := string(it.Value())
+
+		require.Equal(t, nums[i], string(key.UserKey))
+		require.Equal(t, nums[i], string(value))
+		i++
+	}
+	require.Equal(t, num, i)
+	require.Nil(t, it.Err())
+}
```
coocood: We need to test that `SeekToFirst` called more than once.

rocksdb/sst_file_iterator.go
```diff
@@ -0,0 +1,196 @@
...
+	if it.dataBuf, err = it.decompressBlock(it.dataBuf); err != nil {
+		return err
+	}
+	it.dataBlockIter.Reset(it.dataBuf)
+
+	return nil
+}
+
+func (it *SstFileIterator) changeBufSize(sz uint64) {
+	if uint64(len(it.dataBuf)) < sz {
```
coocood: Use `cap` instead of `len` can save some memory allocation.

rocksdb/sst_file_iterator.go
```diff
@@ -0,0 +1,196 @@
...
+	trailerPos := len(data) - blockTrailerSize
+
+	blkData := data[:trailerPos]
+	compressTp := CompressionType(data[trailerPos])
+
+	switch it.checksumType {
+	case ChecksumCRC32:
+		crc := newCrc32()
+		crc.Write(data[:trailerPos])
+		crc.Write([]byte{data[trailerPos]})
```
coocood: Why not ```crc.Write(data[:trailerPos+1]```?

rocksdb/sst_file_iterator.go
bobotu: In this case `len(buf)` always equals to `cap(buf)`.

rocksdb/sst_file_iterator.go
bobotu: Did you mean `crc.Write(data[trailerPos:trailerPos+1])`?

coocood: Then we can call `Write` only once.

coocood:
LGTM

PR #110, opened 2019-01-26
extract mvcccodec package

coocood:
The functions in `types.go` will be used by raftstore snapshot which is in a different package.
extract it to avoid cyclic dependency.


coocood:
@bobotu PTAL

tikv/db_reader.go
```diff
@@ -8,6 +8,7 @@ import (
 	"github.com/coocood/badger"
 	"github.com/coocood/badger/y"
 	"github.com/juju/errors"
+	"github.com/ngaut/unistore/tikv/mvcccodec"
```
ngaut: I prefer rename mvcccodec to mvcc. let's keep it simple and clear.

ngaut:
Rest LGTM.

PR #111, opened 2019-01-28
port TiKV Snapshot module

coocood:
ported and passed the united test in TiKV.


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/engine.go
```diff
@@ -3,9 +3,17 @@ package raftstore
...
 	"github.com/coocood/badger"
 	"github.com/golang/protobuf/proto"
+	"github.com/ngaut/unistore/lockstore"
 	"github.com/pingcap/errors"
 )
 
+type DBSnapshot struct {
+	db            *badger.DB
+	txn           *badger.Txn
+	lockStore     *lockstore.MemStore
```
zhangjinpeng1987: Do we need to keep the consistency between badger.DB and MemStore?

tikv/raftstore/keys.go
```diff
@@ -101,10 +103,41 @@ func RegionStateKey(regionID uint64) []byte {
 	return key
 }
 
+func ValidateDataKey(key []byte) bool {
+	return len(key) > 0 && key[0] == DataPrefix
+}
+
+func DataKey(key []byte) []byte {
+	v := make([]byte, 0, 1+len(key))
```
zhangjinpeng1987: Is it possible to reuse memory space with key in golang?

tikv/raftstore/keys.go
```diff
@@ -101,10 +103,41 @@ func RegionStateKey(regionID uint64) []byte {
...
+}
+
+func OriginKey(key []byte) []byte {
+	y.AssertTruef(ValidateDataKey(key), "invalid data key %v", key)
+	return key[1:]
+}
+
+func DataEndKey(key []byte) []byte {
+	if len(key) == 0 {
+		return DataMinKey
```
zhangjinpeng1987: I think it should be `DataMaxKey`.

tikv/raftstore/engine.go
coocood: Yes, we should handle it later.

tikv/raftstore/keys.go
coocood: Yes, just pass a []byte as an argument then appends to it.

we can optimize it later.

coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+	}
+}
+
+func SnapKeyFromSnap(snap *raftpb.Snapshot) (*SnapKey, error) {
+	data := new(rspb.RaftSnapshotData)
+	err := data.Unmarshal(snap.Data)
+	if err != nil {
+		return nil, err
+	}
+	return &SnapKey{
```
zhangjinpeng1987: Better to reuse `SnapKeyFromRegionSnap`?

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+		RegionID: data.Region.Id,
+		Term:     snap.Metadata.Term,
+		Index:    snap.Metadata.Index,
+	}, nil
+}
+
+type SnapStatistics struct {
+	Size    uint64
+	KVCount int
```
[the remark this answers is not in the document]
coocood: We rarely use uint in Go. `int` is more convenient to use.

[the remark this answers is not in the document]
zhangjinpeng1987: Got it.

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+}
+
+type ApplyOptions struct {
+	DBSnap    *DBSnapshot
```
zhangjinpeng1987: In TiKV, this field is RocksDB::DB, DBSnapshot is a struct combine badger and memstore, i prefer rename DBSnapshot to UniDB.

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+	return false
+}
+
+func genSnapshotMeta(cfFiles []*CFFile) (*rspb.SnapshotMeta, error) {
+	cfMetas := make([]*rspb.SnapshotCFFile, 0, len(snapshotCFs))
+	for _, cfFile := range cfFiles {
+		var found bool
+		for _, snapCF := range snapshotCFs {
+			if snapCF == cfFile.CF {
+				found = true
```
zhangjinpeng1987: break

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+type MetaFile struct {
+	Meta *rspb.SnapshotMeta
+	Path string
+	File *os.File
+
+	// for writing snapshot
+	TmpPath string
+}
+
+var _ Snapshot = new(Snap)
```
zhangjinpeng1987: What it the purpose for this anonymous variable.

coocood: To make compiler check that the `Snap` implements `Snapshot` interface.

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+			CF:        cf,
+			Path:      path,
+			TmpPath:   tmpPath,
+			ClonePath: clonePath,
+		}
+		cfFiles = append(cfFiles, cfFile)
+	}
+	metaFileName := fmt.Sprintf("%s%s", prefix, metaFileSuffix)
+	metaFilePath := filepath.Join(dir, metaFileName)
+	metaTmpPath := metaFilePath + metaFileSuffix
```
zhangjinpeng1987: s/metaFileSuffix/tmpFileSuffix

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+			// this is checked when loading the snapshot meta.
+			continue
+		}
+		if plainFileUsed(cfFile.CF) {
+			err := checkFileSizeAndChecksum(cfFile.Path, cfFile.Size, cfFile.Checksum)
+			if err != nil {
+				return err
+			}
+		} else {
+			// TODO: prepare and validate for ingestion
```
zhangjinpeng1987: Use sstReader or badger support Ingestion?

coocood: It will be supported later.

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+		}
+	}
+	return nil
+}
+
+func (s *Snap) saveCFFiles() error {
+	for _, cfFile := range s.CFFiles {
+		if !plainFileUsed(cfFile.CF) && cfFile.KVCount > 0 {
```
zhangjinpeng1987: Also need to close SstWrite if KVcount == 0?

coocood: tikv didn't do that, jus
```
 let _ = cf_file.sst_writer.take().unwrap();
```

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+}
+
+func (s *Snap) Save() error {
+	log.Debugf("saving to %s", s.MetaFile.Path)
+	for _, cfFile := range s.CFFiles {
+		if cfFile.Size == 0 {
+			// skip empty cf file.
+			continue
+		}
+		// Check each cf file has been fully written, and the checksum matches.
```
zhangjinpeng1987: Don't need to flush explicitly?

tikv/raftstore/snap.go
```diff
@@ -0,0 +1,834 @@
...
+			if batch.size >= opts.BatchSize {
+				err = batch.WriteToDB(opts.DBSnap.db)
+				if err != nil {
+					return err
+				}
+				batch = new(WriteBatch)
+			}
+		case applySnapTypeLock:
+			opts.DBSnap.lockStore.Insert(item.key, item.val)
+		case applySnapTypeRollback:
```
zhangjinpeng1987: `Rollback` only exist between two `unistore`?

tikv/raftstore/snap.go
coocood: No, the `*os.FIle` doesn't even have the `flush` method.

tikv/raftstore/snap.go
zhangjinpeng1987: My mistake, please ignore.

tikv/raftstore/snap_applier.go
```diff
@@ -0,0 +1,251 @@
...
+
+func (ai *snapApplier) loadOldValue(lock *mvcc.MvccLock) error {
+	if lock.Op == byte(kvrpcpb.Op_Lock) {
+		return nil
+	}
+	for bytes.Equal(ai.curLockKey, ai.curWriteKey) {
+		var err error
+		writeVal := decodeWriteCFValue(y.SafeCopy(nil, ai.writeCFIterator.Value()))
+		if writeVal.writeType == byte(kvrpcpb.Op_Rollback) {
+			// Skip rollback entry, find the next write key.
```
zhangjinpeng1987: How about if there is a prewrite in default cf and a rollback in write cf?

coocood: The key in the default CF will be deleted when the rollback is written.

zhangjinpeng1987: Also need to skip the related value in default cf.

coocood: I mean the default CF key is already deleted when the rollback is written in TiKV.

tikv/raftstore/snap_builder.go
```diff
@@ -0,0 +1,231 @@
...
+	b.endKey = rawRegionKey(region.EndKey)
+	b.txn = db.NewTransaction(false)
+	b.dbIterator = b.txn.NewIterator(badger.DefaultIteratorOptions)
+	startKey := rawRegionKey(region.StartKey)
+	b.dbIterator.Seek(startKey)
+	if b.dbIterator.Valid() && !b.reachEnd(b.dbIterator.Item().Key()) {
+		b.curDBKey = b.dbIterator.Item().Key()
+	}
+	b.dbOldIterator = b.txn.NewIterator(badger.DefaultIteratorOptions)
+	b.dbOldIterator.Seek(mvcc.EncodeOldKey(startKey, math.MaxUint64))
```
zhangjinpeng1987: I don't understand how to distinguish old iterator and latest iterator since they are built with the same API NewIterator.

tikv/raftstore/snap_builder.go
```diff
@@ -0,0 +1,231 @@
...
+	b.lockIterator.Seek(startKey)
+	if b.lockIterator.Valid() && !b.reachEnd(b.lockIterator.Key()) {
+		b.curLockKey = b.lockIterator.Key()
+	}
+	lockCFFile := cfFiles[lockCFIdx].File
+	if lockCFFile == nil {
+		return nil, errors.New("lock CF file is nil")
+	}
+	b.lockCFWriter = cfFiles[lockCFIdx].File
+	b.defaultCFWriter = cfFiles[defaultCFIdx].SstWriter
```
zhangjinpeng1987: Also need to verify default cf sstWriter and write cf sstWriter like lock cf file?

tikv/raftstore/snap_builder.go
```diff
@@ -0,0 +1,231 @@
...
+		return err
+	}
+	b.cfFiles[writeCFIdx].KVCount++
+	b.size += len(writeCFKey) + len(encodedWriteCFVal)
+	b.kvCount++
+	return nil
+}
+
+func isLatestKeyAndOldKeySame(lastestKey, oldKey []byte) bool {
+	return len(oldKey)-len(lastestKey) == 8 && bytes.Equal(lastestKey[1:], oldKey[1:len(lastestKey)])
```
zhangjinpeng1987: There is a potential optimization that compare the last byte after length match, because these two keys may have several common beginning bytes.

tikv/raftstore/snap_builder.go
coocood: They seek different ranges.

tikv/raftstore/snap_builder.go
coocood: We can do this optimization when we see it in profile.

tikv/raftstore/snap_applier.go
zhangjinpeng1987: OK

tikv/raftstore/snap_builder.go
zhangjinpeng1987: Got it, so the old values has a different prefix?

zhangjinpeng1987:
LGTM

PR #116, opened 2019-02-13
port raftstore SnapManager

coocood:


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/snap_manager.go
```diff
@@ -0,0 +1,355 @@
...
+	} else if _, ok := sm.registry[key]; ok {
+		log.Infof("skip to delete %s since it's registered.", snapshot.Path())
+		return false
+	}
+	snapshot.Delete()
+	return true
+}
+
+type SnapManagerBuilder struct {
+	maxTotalSize        uint64
```
zhangjinpeng1987: The max_write_bytes_per_sec is missing?

coocood: badger doesn't support the limiter feature yet.

tikv/raftstore/snap_manager.go
```diff
@@ -0,0 +1,355 @@
...
+	limiter      *IOLimiter
+	MaxTotalSize uint64
+}
+
+func NewSnapManager(path string, ch chan<- *Msg) *SnapManager {
+	return new(SnapManagerBuilder).Build(path, ch)
+}
+
+func (sm *SnapManager) init() error {
+	fi, err := os.Stat(sm.base)
```
zhangjinpeng1987: Should acquire write lock to prevent multiple threads call init at the same time.

coocood: I suppose SnapManger.init() would only be called once.

tikv/raftstore/snap_manager.go
```diff
@@ -0,0 +1,355 @@
...
+		}
+		return keyI.RegionID < keyJ.RegionID
+	})
+	return results, nil
+}
+
+func (sm *SnapManager) HasRegistered(key SnapKey) bool {
+	sm.registryLock.RLock()
+	_, ok := sm.registry[key]
+	sm.registryLock.RUnlock()
```
zhangjinpeng1987: Why not use defer?

tikv/raftstore/snap_manager.go
```diff
@@ -0,0 +1,355 @@
...
+	return ok
+}
+
+func (sm *SnapManager) GetTotalSnapSize() uint64 {
+	return uint64(atomic.LoadInt64(sm.snapSize))
+}
+
+func (sm *SnapManager) GetSnapshotForBuilding(key SnapKey, bundle *DBBundle) (Snapshot, error) {
+	if sm.GetTotalSnapSize() > sm.MaxTotalSize {
```
zhangjinpeng1987: s/if/while

tikv/raftstore/snap_manager.go
```diff
@@ -0,0 +1,355 @@
...
+			return err
+		}
+		snaps = append(snaps, snapWithModTime{key: idleSnap.SnapKey, snap: snap, modTime: fi.ModTime()})
+	}
+	sort.Slice(snaps, func(i, j int) bool {
+		return snaps[i].modTime.Before(snaps[j].modTime)
+	})
+	for sm.GetTotalSnapSize() > sm.MaxTotalSize {
+		if len(snaps) == 0 {
+			return errors.New("too many snapshots")
```
zhangjinpeng1987: Should collect several rounds?

zhangjinpeng1987: There maybe some old snapshots can be deleted in the next round.

tikv/raftstore/snap_manager.go
zhangjinpeng1987: OK

tikv/raftstore/snap_manager.go
zhangjinpeng1987: Got it.

tikv/raftstore/snap_manager.go
coocood: I think It's ok to not defer if there is just one line of code.

tikv/raftstore/snap_manager.go
coocood: `for` loop is done in `deleteOldIdleSnaps`.
Here we just need `if`.

tikv/raftstore/snap_manager.go
coocood: The passed duration is very short, it's unlikely that `ListIdleSnap` again can avoid this error.
And the code is simpler when we only call `ListIdleSnap` once.

Issue #113, opened 2019-01-30
port coprocessor unit tests from tikv to unistore


coocood:
port the test code in
https://github.com/tikv/tikv/blob/master/src/coprocessor/endpoint.rs
to 'unistore/tikv' package.

coocood:
@xiaobogaga

PR #117, opened 2019-02-15
port point_get test from table_scan.rs to unistore (#113)

xiaobogaga:
port the point_get test from [table_scan.rs](https://github.com/tikv/tikv/blob/master/src/coprocessor/dag/executor/table_scan.rs) to cop_handler_test.go in unistore

### DIFF
--- a/tikv/cop_handler_test.go
+++ b/tikv/cop_handler_test.go
@@ -1,0 +1,363 @@
+package tikv
+
+import (
+	"fmt"
+	"github.com/coocood/badger"
+	"github.com/pingcap/kvproto/pkg/coprocessor"
+	"github.com/pingcap/kvproto/pkg/kvrpcpb"
+	"github.com/pingcap/kvproto/pkg/metapb"
+	"github.com/pingcap/parser/mysql"
+	"github.com/pingcap/tidb/kv"
+	"github.com/pingcap/tidb/sessionctx/stmtctx"
+	"github.com/pingcap/tidb/util/codec"
+	"golang.org/x/net/context"
+	"math"
+	"time"
+
+	// remember the tablecodec has two different versions,
+	// one is on the tibd/tablecodec/origin/tablecodec.go
+	// and another is on tidb/tablecodec/tablecodec.go,
+	// this must changes with the isShardingEnabled parameter.
+	"github.com/pingcap/tidb/tablecodec/origin"
+	"github.com/pingcap/tidb/types"
+	"github.com/pingcap/tipb/go-tipb"
+	"github.com/stretchr/testify/require"
+	"os"
+	"sync"
+	"testing"
+)
+
+const (
+	keyNumber = 3
+	TableId   = 0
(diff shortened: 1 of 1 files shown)


tikv/cop_handler_test.go
```diff
@@ -0,0 +1,363 @@
...
+	data = PrepareTestTableData(t, keyNumber, TableId)
+	store, err = NewTestStore()
+	require.Nil(t, err)
+	errors = store.InitTestData(data.encodedKVDatas)
+	require.Nil(t, errors)
+	// point get should return nothing when handle is math.MinInt64
+	chunks, err = executeAPointGet(data, store, math.MinInt64)
+	require.Nil(t, err)
+	require.Equal(t, len(chunks), 0)
+	// point get should return one row when handler = 0
```
coocood: s/handler/handle

tikv/cop_handler_test.go
```diff
@@ -0,0 +1,363 @@
...
+
+func TestPointGet(t *testing.T) {
+	var (
+		data   *data
+		store  *TestStore
+		err    error
+		errors []error
+		chunks []tipb.Chunk
+		row    []types.Datum
+		l      int
```
coocood: Better define the local variable where it's first used.
And what does 'l' mean?

xiaobogaga: oops, my fault, I would change the name of this parameter and define it as a local variable

tikv/cop_handler_test.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+func executeAPointGet(data *data, store *TestStore, handler int64) (chunks []tipb.Chunk, err error) {
+	// we would prepare the dagRequest and try build a executor as follows,
+	// the reason that we don't make a tableScanExecutor directly here is
+	// it would be replaced by closureExecutor in the future, so, for compatibility,
+	// we just build a dagReq, and call buildDAGExecutor to build
+	// the corresponding executor.
+	aRange := getTestPointRange(TableId, handler)
+	dagReq := &tipb.DAGRequest{
```
coocood: If we extract the DAG builder, we can use it for other tests.

tikv/cop_handler_test.go
xiaobogaga: did you mean the comment `point get should return one row when handler = 0` should replaced by `point get should return one row when handle = 0` ?

coocood: swap the typo `handler` with `handle`

xiaobogaga: ok

tikv/cop_handler_test.go
xiaobogaga: Yes, but I don't know what it is like for extracting DAG builder, for example, what parameters does it needs, so it might be better that do that when we needs more test using DAG builder, and I can figure out what it likes to extract dag builder. What do you think?

coocood:
LGTM

PR #119, opened 2019-02-18
port FSM router and batch system

coocood:


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+		}
+		prevStatus = atomic.LoadInt64(&mb.state.status)
+		if prevStatus == notifyStateDrop {
+			atomic.SwapPointer(&mb.state.data, nil)
+		}
+	}
+	panic(fmt.Sprintf("invalid release state %d", prevStatus))
+}
+
+func (mb *mailbox) len() int {
```
zhangjinpeng1987: Is there `inline` key words?

coocood: Go compiler will inline simply code automatically.

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+	}
+}
+
+func (mb *mailbox) send(msg *Msg, scheduler fsmScheduler) {
+	mb.sender <- msg
+	mb.notify(scheduler)
+}
+
+func (mb *mailbox) close() {
+	prevState := atomic.SwapInt64(&mb.state.status, notifyStateDrop)
```
zhangjinpeng1987: Do we need to close mb.sender?

coocood: We don't need to and we cannot close it, because the sender will be sent by multiple threads.

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	old := r.normals[addr]
+	r.normals[addr] = mailbox
+	if old != nil {
+		old.close()
+	}
+}
+
+type addrWithMailbox struct {
```
zhangjinpeng1987: s/addrWithMailbox/addrMailboxPair

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+}
+
+func (r *router) broadcastShutdown() {
+	log.Info("broadcasting shutdown")
+	r.mu.Lock()
+	defer r.mu.Unlock()
+	for addr, mb := range r.normals {
+		log.Debugf("region %d shutdown mailbox", addr)
+		mb.close()
+	}
```
zhangjinpeng1987: Need to close the controlBox?

zhangjinpeng1987: Also need to clean r.normals.

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+		l++
+	}
+	return l
+}
+
+func (b *batch) isEmpty() bool {
+	return b.len() == 0
+}
+
+func (b *batch) clear() {
```
zhangjinpeng1987: Should we need to add a shrink function to reclaim the memory?

coocood: No the operation `b.normals = b.normals[:0]` reuses the memory.

tikv/raftstore/fsm.go
```diff
@@ -0,0 +1,556 @@
...
+
+type poller struct {
+	router       *router
+	fsmReceiver  <-chan *Msg
+	handler      pollHandler
+	maxBatchSize int
+}
+
+func (p *poller) fetchBatch(batch *batch, maxSize int) {
+	msg, ok := <-p.fsmReceiver
```
zhangjinpeng1987: This rely on an assumption that there must be some messages in the channel. The latest TiKV has some changes for this function, please sync it.

PR #121, opened 2019-02-20
Raftstore peer

zhangjinpeng1987:
Last part of peer.rs.


zhangjinpeng1987:
@coocood PTAL

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+			}
+			apply := &ApplyTask{
+				RegionId: p.regionId,
+				Term:     p.Term(),
+				Entries:  committedEntries,
+			}
+			ctx.applyRouter.ScheduleTask(p.regionId, apply)
+		}
+
+		p.ApplyReads(ctx, ready)
```
coocood: Should move out of the `else` block.

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+	readyReadCount := p.RaftGroup.Raft.ReadyReadCount()
+
+	if pendingReadCount == lastPendingReadCount && readyReadCount == lastReadyReadCount {
+		// The message gets dropped silently, can't be handled anymore.
+		NotifyStaleReq(p.Term(), cb)
+		return false
+	}
+
+	cmds := make([]*ReqCbPair, 0, 1)
+	cmds = append(cmds, &ReqCbPair{req, cb})
```
coocood: This is simpler.
```
cmds := []*ReqCbPair{&ReqCbPair{req, cb}}
```

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+			return nil, err
+		}
+		ctx.insert(ProposalContext_PrepareMerge)
+	}
+
+	return ctx, nil
+}
+
+func (p *Peer) ProposeNormal(pollCtx *PollContext, req *raft_cmdpb.RaftCmdRequest) (uint64, error) {
+	if p.PendingMergeState != nil && (req.AdminRequest == nil || req.AdminRequest.CmdType != raft_cmdpb.AdminCmdType_RollbackMerge) {
```
coocood: We can call `req.AdminRequest.GetCmdType()` directly without checking if `AdminRequest` is nil.

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+		return 0, err
+	}
+
+	if uint64(len(data)) > pollCtx.Cfg.RaftEntryMaxSize {
+		log.Errorf("entry is too large, entry size %v", len(data))
+		return 0, &ErrRaftEntryTooLarge{RegionId: p.regionId, EntrySize: uint64(len(data))}
+	}
+
+	proposeIndex := p.nextProposalIndex()
+	// Todo: after the raft implement propose with context, chagne to:
```
coocood: s/Todo/TODO

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+		return 0, &ErrNotLeader{RegionId: p.regionId}
+	}
+
+	return proposeIndex, nil
+}
+
+// Return true if the transfer leader request is accepted.
+func (p *Peer) ProposeTransferLeader(ctx *PollContext, req *raft_cmdpb.RaftCmdRequest, cb Callback) bool {
+	transferLeader := getTransferLeaderCmd(req)
+	if transferLeader == nil {
```
coocood: We can remove this check, it will panic when we call `transferLeaer.Peer`.

tikv/raftstore/peer.go
```diff
@@ -992,3 +1028,930 @@ func (p *Peer) sendRaftMessage(msg eraftpb.Message, trans chan<- *rspb.RaftMessa
...
+		return 0, err
+	}
+
+	data, err := req.Marshal()
+	if err != nil {
+		return 0, err
+	}
+
+	changePeer := GetChangePeerCmd(req)
+	if changePeer == nil {
```
coocood: ditto

tikv/raftstore/peer.go
zhangjinpeng1987: No, if `AdminRequest` is nil, here will panic, but TiKV will not panic.

coocood:
LGTM

PR #122, opened 2019-02-20
raftstore error

zhangjinpeng1987:
Add raftstore errors


tikv/raftstore/error.go
```diff
@@ -0,0 +1,117 @@
...
+	EntrySize uint64
+}
+
+func (e *ErrRaftEntryTooLarge) Error() string {
+	return fmt.Sprintf("raft entry too large, region_id: %v, len: %v", e.RegionId, e.EntrySize)
+}
+
+func RaftstoreErrToPbError(e error) *errorpb.Error {
+	ret := new(errorpb.Error)
+	if notLeader, ok := errors.Cause(e).(*ErrNotLeader); ok {
```
coocood: How about switch case?
```
switch x := errors.Cause(e).(type) {
case *ErrNotLeader:
    ret.NotLeader = &errorpb.NotLeader{ RegionId: x.RegionId, Leader: x.Leader }
case *ErrRegionNotFound:
    ...
}

```

zhangjinpeng1987: Done

coocood:
LGTM

PR #123, opened 2019-02-25
use etcd-io instead of coreos

zhangjinpeng1987:

TiKV has some changes for raft, for example make some private members public, add context for propose. We are going to keep consistent with tikv, this is the first step.


coocood:
LGTM

PR #124, opened 2019-02-25
move pd client to pd package

coocood:
We will later add more PD features in it.

### DIFF
--- a/pd/client.go
+++ b/pd/client.go
@@ -11,7 +11,7 @@
 // See the License for the specific language governing permissions and
 // limitations under the License.
 
-package tikv
+package pd
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


pd/client.go
```diff
@@ -55,13 +55,18 @@ type client struct {
...
+	regionCh                 chan regInfoSizeHintPair
 
 	wg     sync.WaitGroup
 	ctx    context.Context
 	cancel context.CancelFunc
 }
 
+type regInfoSizeHintPair struct {
+	info     *metapb.Region
+	sizeHint int64
```
zhangjinpeng1987: What does `hint` means here?

coocood: approximate size, used to report to PD.

PR #130, opened 2019-02-26
impl BatchCommands func

bobotu:
A rough implement of batch gRPC messages.


tikv/server_batch.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		h.responseBuffer.ids, batchResp.RequestIds = batchResp.RequestIds, h.responseBuffer.ids
+		h.responseBuffer.resps, batchResp.Responses = batchResp.Responses, h.responseBuffer.resps
+		h.responseBuffer.Unlock()
+
+		if err := h.stream.Send(batchResp); err != nil {
+			return err
+		}
+		batchResp.Responses = batchResp.Responses[:0]
+		batchResp.RequestIds = batchResp.RequestIds[:0]
```
coocood: If we don't wait, then the batchResponse may only contain one response.

coocood:
LGTM

PR #131, opened 2019-02-27
[peer storage] post ready

zhangjinpeng1987:


tikv/raftstore/keys.go
```diff
@@ -122,6 +123,22 @@ func OriginKey(key []byte) []byte {
 	return key[1:]
 }
 
+// Get the `start_key` of current region in encoded form.
+func EncStartKey(region *metapb.Region) []byte {
+	// only initialized region's start_key can be encoded, otherwise there must be bugs
+	// somewhere.
+	y.AssertTruef(len(region.Peers) > 0, "region is not initialized")
```
coocood: I prefer `y.Assert` than `y.AssertTruef` for better performance.

tikv/raftstore/peer_storage.go
```diff
@@ -19,15 +19,32 @@ import (
...
+	SnapState_Relax SnapStateType = 0 + iota
 	SnapState_Generating
 	SnapState_Applying
 	SnapState_ApplyAborted
 )
 
+type SnapState struct {
+	StateType SnapStateType
+	Status    *JobStatus
+	Receiver  chan<- eraftpb.Snapshot
```
coocood: better use pointer type in chan.

tikv/raftstore/peer_storage.go
```diff
@@ -134,6 +151,32 @@ func (ec *EntryCache) compactTo(idx uint64) {
...
+	///
+	/// The deletion may and may not succeed.
+	RegionTaskType_Destroy
+)
+
+type RegionTask struct {
+	RegionId uint64
+	TaskType RegionTaskType
+	Notifier chan<- eraftpb.Snapshot
+	Status   *JobStatus
```
coocood: We don't need to use pointer.

zhangjinpeng1987: In TiKV, it is a atomic usize shared by multiple threads.

coocood: I know, we can use int64 type to do atomic operations.

tikv/raftstore/peer_storage.go
```diff
@@ -134,6 +151,32 @@ func (ec *EntryCache) compactTo(idx uint64) {
...
+	/// Destroy data between [start_key, end_key).
+	///
+	/// The deletion may and may not succeed.
+	RegionTaskType_Destroy
+)
+
+type RegionTask struct {
+	RegionId uint64
+	TaskType RegionTaskType
+	Notifier chan<- eraftpb.Snapshot
```
coocood: pointer

tikv/raftstore/peer_storage.go
zhangjinpeng1987: Yes, JobStatus is already int/int64.

zhangjinpeng1987: And `JobStatus` shared by `RegionTask` and `SnapState` which in different threads.

coocood:
LGTM

PR #134, opened 2019-02-27
refactor writeDBWorker and server batch to use channel instead of mutex.

coocood:
I recently did a microbenchmark for the two approaches of batching, the difference is very small.
Changing mutex to channel can simplify the code, easier to understand.


coocood:
@bobotu PTAL

PR #135, opened 2019-02-28
remove etcd/raft dependency

zhangjinpeng1987:
Etcd use its own proto file, there are some conflicts with kvproto/eraftpb.


coocood:
We can put the raft code under the repository root to simplify import path.

coocood:
LGTM

PR #136, opened 2019-03-01
use MultiGet in Prewrite

coocood:


ngaut:
LGTM, please merge dependent PR in https://github.com/coocood/badger/pull/80 first.

PR #137, opened 2019-03-04
check region epoch

zhangjinpeng1987:

### DIFF
--- a/tikv/raftstore/util.go
+++ b/tikv/raftstore/util.go
@@ -1,15 +1,37 @@
 package raftstore
 
 import (
-	"time"
+	"bytes"
+	"fmt"
+	"github.com/ngaut/log"
+	"github.com/pingcap/kvproto/pkg/eraftpb"
+	"github.com/pingcap/kvproto/pkg/metapb"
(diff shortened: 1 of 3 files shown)


tikv/raftstore/util.go
```diff
@@ -231,3 +253,67 @@ func U64ToTime(u uint64) time.Time {
...
+func CheckRegionEpoch(req *raft_cmdpb.RaftCmdRequest, region *metapb.Region, includeRegion bool) error {
+	checkVer, checkConfVer := false, false
+	if req.AdminRequest == nil {
+		checkVer = true
+	} else {
+		switch req.AdminRequest.CmdType {
+		case raft_cmdpb.AdminCmdType_CompactLog, raft_cmdpb.AdminCmdType_InvalidAdmin,
+			raft_cmdpb.AdminCmdType_ComputeHash, raft_cmdpb.AdminCmdType_VerifyHash:
+			{
+			}
```
coocood: We don't need the two braces.

coocood:
LGTM

PR #138, opened 2019-03-05
init peer

zhangjinpeng1987:


tikv/raftstore/peer.go
```diff
@@ -0,0 +1,375 @@
...
+	PendingMergeState *rspb.MergeState
+	leaderMissingTime *time.Time
+	leaderLease *Lease
+
+	// If a snapshot is being applied asynchronously, messages should not be sent.
+	pendingMessages []eraftpb.Message
+	PendingMergeApplyResult *WaitApplyResultStat
+	PeerStat *PeerStat
+
+	applyRouter router
```
coocood: This field is not present in TikV.

tikv/raftstore/peer.go
```diff
@@ -0,0 +1,375 @@
...
+
+type DestroyPeerJob struct {
+	Initialized bool
+	AsyncRemove bool
+	RegionId uint64
+	Peer *metapb.Peer
+}
+
+type Peer struct {
+	Cfg *Config
```
coocood: TiKV doesn't have this field.

tikv/raftstore/peer.go
```diff
@@ -0,0 +1,375 @@
...
+	raftCfg := &raft.Config {
+		ID: peer.GetId(),
+		ElectionTick: cfg.RaftElectionTimeoutTicks,
+		HeartbeatTick: cfg.RaftHeartbeatTicks,
+		MaxSizePerMsg: cfg.RaftMaxSizePerMsg,
+		MaxInflightMsgs: cfg.RaftMaxInflightMsgs,
+		Applied: appliedIndex,
+		CheckQuorum: true,
+		PreVote: cfg.Prevote,
+		Storage: ps,
```
coocood: Missing `skip_bcast_commit`?

tikv/raftstore/peer.go
```diff
@@ -0,0 +1,375 @@
...
+		RaftGroup: raftGroup,
+		peerStorage: ps,
+		proposals: newProposalQueue(),
+		applyProposals: make([]*Proposal, 0),
+		pendingReads: NewReadIndexQueue(),
+		peerCache: make(map[uint64]*metapb.Peer),
+		PeerHeartbeats: make(map[uint64]time.Time),
+		PeersStartPendingTime: make(map[uint64]time.Time),
+		RecentAddedPeer: NewRecentAddedPeer(uint64(cfg.RaftRejectTransferLeaderDuration.Seconds())),
+		SizeDiffHint: 0,
```
coocood: I think we can remove those zero value fields.

tikv/raftstore/peer.go
zhangjinpeng1987: Will add it later, not this pr.

coocood:
LGTM

PR #145, opened 2019-03-10
propose with context

zhangjinpeng1987:

Propose with context after raft lib has implemented the feature.


coocood:
LGTM

PR #147, opened 2019-03-12
port PeerFsm step 1

coocood:
The ticker is different from TiKV but does the same work.

### DIFF
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -1,0 +1,643 @@
+package raftstore
+
+import (
+	"time"
+
+	"github.com/coocood/badger/y"
+	"github.com/ngaut/log"
+	"github.com/pingcap/errors"
+	"github.com/pingcap/kvproto/pkg/import_sstpb"
+	"github.com/pingcap/kvproto/pkg/metapb"
+	"github.com/pingcap/kvproto/pkg/pdpb"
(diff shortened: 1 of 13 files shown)


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/fsm_peer.go
```diff
@@ -0,0 +1,643 @@
...
+	// a value that is larger than last index.
+	if d.peer.IsApplyingSnapshot() || d.peer.HasPendingSnapshot() {
+		// need to check if snapshot is applied.
+		d.hasReady = true
+		d.ticker.schedule(PeerTickRaft)
+		return
+	}
+	// TODO: make Tick returns bool to indicate if there is ready.
+	d.peer.RaftGroup.Tick()
+	if d.peer.RaftGroup.HasReady() {
```
zhangjinpeng1987: `d.hasRead = d.peer.RaftGroup.HasReady()`

tikv/raftstore/fsm_peer.go
```diff
@@ -0,0 +1,643 @@
...
+			d.onClearRegionSize()
+		case MsgTypeStart:
+			d.start()
+		case MsgTypeNoop:
+		}
+	}
+}
+
+func (d *peerFsmDelegate) onTick(tick PeerTick) {
+	d.ticker.step()
```
zhangjinpeng1987: Missing `if d.fsm.stopped()`?

PR #149, opened 2019-03-13
port PeerFsm step 2

coocood:
Refactor `Msg` to use `interface` instead of multiple fields.
Since the struct is small, only 32 bytes, we can use value instead of pointer.
Move some code from peer to fsm_apply and fsm_store.

### DIFF
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -525,7 +525,7 @@ func (d *peerFsmDelegate) needGCMerge(msg *rspb.RaftMessage) (bool, error) {
 
 func (d *peerFsmDelegate) handleGCPeerMsg(msg *rspb.RaftMessage) {
 	fromEpoch := msg.RegionEpoch
-	if !IsEpochStale(fromEpoch, d.peer.Region().RegionEpoch) {
+	if !IsEpochStale(d.peer.Region().RegionEpoch, fromEpoch) {
 		return
 	}
 	if !PeerEqual(d.peer.Peer, msg.ToPeer) {
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -1,6 +1,9 @@
 package raftstore
 
 import (
+	"bytes"
+	"encoding/binary"
+	"fmt"
 	"time"
 
 	"github.com/coocood/badger/y"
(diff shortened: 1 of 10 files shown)


tikv/raftstore/fsm_peer.go
```diff
@@ -464,19 +474,165 @@ func (d *peerFsmDelegate) validateRaftMessage(msg *rspb.RaftMessage) bool {
...
 }
 
 func (d *peerFsmDelegate) needGCMerge(msg *rspb.RaftMessage) (bool, error) {
 	return false, nil // TODO: merge func
 }
 
 func (d *peerFsmDelegate) handleGCPeerMsg(msg *rspb.RaftMessage) {
-	// TODO: stub
+	fromEpoch := msg.RegionEpoch
+	if !IsEpochStale(fromEpoch, d.peer.Region().RegionEpoch) {
```
zhangjinpeng1987: Seems `fromEpoch` should be the second parameter?

tikv/raftstore/fsm_peer.go
```diff
@@ -464,19 +474,165 @@ func (d *peerFsmDelegate) validateRaftMessage(msg *rspb.RaftMessage) bool {
...
+	meta := d.ctx.storeMeta
+	if !RegionEqual(meta.regions[d.regionID()], d.region()) {
+		if !d.peer.isInitialized() {
+			log.Infof("%s stale delegate detected, skip", d.tag())
+			return &key, nil
+		} else {
+			panic(fmt.Sprintf("%s meta corrupted %s != %s", d.tag(), meta.regions[d.regionID()], d.region()))
+		}
+	}
+	existOverlapRegion := d.findOverlapRegion(meta, snapRegion)
```
zhangjinpeng1987: I don't see this logic in TiKV repo.

tikv/raftstore/fsm_store.go
```diff
@@ -1,5 +1,88 @@
...
+		FromPeer:    ClonePeer(fromPeer),
+		ToPeer:      ClonePeer(toPeer),
+		RegionEpoch: CloneRegionEpoch(curEpoch),
+	}
+	if targetRegion != nil {
+		gcMsg.MergeTarget = targetRegion
+	} else {
+		gcMsg.IsTombstone = true
+	}
+	err := pc.trans.Send(gcMsg)
```
zhangjinpeng1987: if err:= xxx; err != nil {
}

tikv/raftstore/fsm_peer.go
coocood: This is the corresponding code in tikv
```
        let r = meta
            .region_ranges
            .range((Excluded(enc_start_key(&snap_region)), Unbounded::<Vec<u8>>))
            .map(|(_, &region_id)| &meta.regions[&region_id])
            .take_while(|r| enc_start_key(r) < enc_end_key(&snap_region))
            .skip_while(|r| r.get_id() == region_id)
            .next()
            .map(|r| r.to_owned());
```

zhangjinpeng1987:
LGTM

PR #151, opened 2019-03-19
port PeerFsm step 4

coocood:


tikv/raftstore/util.go
```diff
@@ -360,3 +362,89 @@ func isVoteMessage(msg *eraftpb.Message) bool {
...
+	// info, TiDB updates its region cache and sends requests to TiKV B,
+	// and TiKV B has not applied commit merge yet, since the region epoch in
+	// request is higher than TiKV B, the request must be denied due to epoch
+	// not match, so it does not read on a stale snapshot, thus avoid the
+	// KeyNotInRegion error.
+	if (checkConfVer && fromEpoch.ConfVer != currentEpoch.ConfVer) ||
+		(checkVer && fromEpoch.Version != currentEpoch.Version) {
+		err := &ErrEpochNotMatch{}
+		if includeRegion {
+			err.Regions = []*metapb.Region{region}
```
zhangjinpeng1987: I think you should clone region here?

tikv/raftstore/fsm_peer.go
```diff
@@ -920,113 +917,590 @@ func (d *peerFsmDelegate) onStaleMerge() {
...
+	region := applyResult.Region
+
+	log.Infof("%s snapshot for region %s is applied", d.tag(), region)
+	d.ctx.storeMetaLock.Lock()
+	defer d.ctx.storeMetaLock.Unlock()
+	meta := d.ctx.storeMeta
+	initialized := len(prevRegion.Peers) > 0
+	if initialized {
+		log.Infof("%s region changed from %s -> %s after applying snapshot", d.tag(), prevRegion, region)
+		meta.regionRanges.Delete(prevRegion.EndKey)
```
zhangjinpeng1987: Should check the deleted region's id?

coocood: Then we have to pay the cost of an extra `Get`.
I assume we check it because the `remove` function in TiKV returns the removed item.

tikv/raftstore/util.go
coocood: Right, I `to_own` calls `clone`, I didn't noticed that before.

tikv/raftstore/fsm_peer.go
```diff
@@ -920,113 +917,590 @@ func (d *peerFsmDelegate) onStaleMerge() {
...
+	if err := checkTerm(req, d.peer.Term()); err != nil {
+		return nil, err
+	}
+	err := checkRegionEpoch(req, d.region(), true)
+	if errEpochNotMatching, ok := err.(*ErrEpochNotMatch); ok {
+		// Attach the region which might be split from the current region. But it doesn't
+		// matter if the region is not split from the current region. If the region meta
+		// received by the TiKV driver is newer than the meta cached in the driver, the meta is
+		// updated.
+		siblingRegion := d.findSiblingRegion()
```
zhangjinpeng1987: sliblingRegion here should clone.

tikv/raftstore/fsm_peer.go
```diff
@@ -920,113 +917,590 @@ func (d *peerFsmDelegate) onStaleMerge() {
...
+	defer d.ctx.storeMetaLock.Unlock()
+	meta := d.ctx.storeMeta
+	it := meta.regionRanges.NewIterator()
+	it.Seek(start)
+	valid := it.Valid()
+	if valid && skipFirst {
+		it.Next()
+		valid = it.Valid()
+	}
+	regionID := regionIDFromBytes(it.Value())
```
zhangjinpeng1987: it can be invalid here?

tikv/raftstore/fsm_peer.go
coocood: done

tikv/raftstore/fsm_peer.go
```diff
@@ -920,113 +917,590 @@ func (d *peerFsmDelegate) onStaleMerge() {
...
 func (d *peerFsmDelegate) onCompactionDeclinedBytes(declinedBytes uint64) {
-	// TODO: stub
+	d.peer.CompactionDeclinedBytes += declinedBytes
 }
 
-func (d *peerFsmDelegate) onScheduleHalfSplitRegion(epoch *metapb.RegionEpoch, policy pdpb.CheckPolicy) {
-	// TODO: stub
+func (d *peerFsmDelegate) onScheduleHalfSplitRegion(regionEpoch *metapb.RegionEpoch, policy pdpb.CheckPolicy) {
+	if !d.peer.IsLeader() {
+		log.Warnf("%s not leader, skip", d.tag())
```
zhangjinpeng1987: Miss return here.

tikv/raftstore/fsm_peer.go
```diff
@@ -920,113 +917,590 @@ func (d *peerFsmDelegate) onStaleMerge() {
...
+	}
+	region := d.region()
+	if IsEpochStale(regionEpoch, region.RegionEpoch) {
+		log.Warnf("%s receive a stale halfsplit message", d.tag())
```
zhangjinpeng1987: ditto

PR #155, opened 2019-03-27
Port apply fsm step 1

xiaobogaga:


xiaobogaga:
port data structures, please take a look. @coocood

tikv/import/sst_importer.go
```diff
@@ -0,0 +1,13 @@
+package _import
+
+/// todo, SSTImporter has some methods.
+type SSTImporter struct {
```
coocood: We can skip this feature for now.

xiaobogaga: ok

tikv/raftstore/fsm_apply.go
```diff
@@ -1,12 +1,46 @@
...
 )
 
+type PendingCmd struct {
+	index uint64
+	term  uint64
+	cb    Callback
+}
+
+type PendingCmdQueue struct {
+	normals    *list.List
```
coocood: We can simply use the slice for rust queue.

xiaobogaga: ok.

tikv/raftstore/fsm_apply.go
```diff
@@ -136,6 +206,98 @@ type notifier struct {
...
+	tag             string
+	cfg             *Config
+	coprocessorHost *CoprocessorHost
+	importer        *_import.SSTImporter
+	engines         Engines
+	sender          notifier
+	router          ApplyRouter
+}
+
+type ApplyBatchSystem batchSystem
```
coocood: If you define another type, the original types' method will be lost.

xiaobogaga: Ok, thanks.

tikv/raftstore/fsm_apply.go
```diff
@@ -136,6 +206,98 @@ type notifier struct {
...
+	tp      TaskResType
+	destroy Destroy
+}
+
+type Builder struct {
+	tag             string
+	cfg             *Config
+	coprocessorHost *CoprocessorHost
+	importer        *_import.SSTImporter
+	engines         Engines
```
coocood: use pointer type here

tikv/raftstore/fsm_apply.go
```diff
@@ -136,6 +206,98 @@ type notifier struct {
...
+}
+
+type TaskResType int
+
+const (
+	ApplyTaskResType TaskResType = iota
+	DestroyResType
+)
+
+type TaskRes struct {
```
coocood: We already have `ApplyTaskRes`.

util/time.go
```diff
@@ -0,0 +1,26 @@
...
+
+type InstantType int8
+
+const (
+	Monotonic InstantType = iota
+	MonotonicCoarse
+)
+
+/// todo, Instant has some methods.
+type Instant struct {
```
coocood: We can just use `time.TIme` for rust `Instant`.

util/time.go
```diff
@@ -0,0 +1,26 @@
...
+	it InstantType
+	/// todo, check needed, in TiKV, it uses a time library here.
+	timespec unix.Timespec
+}
+
+/// todo, SlowTimer has some methods.
+type SlowTimer struct {
```
coocood: ditto

util/mod.go
```diff
@@ -0,0 +1,9 @@
+package util
+
+import "container/list"
+
+/// MustConsumerVec is a list with a tag.
+type MustConsumerVec struct {
```
coocood: We can use slice for this type.

xiaobogaga: ok

tikv/raftstore/fsm_apply.go
```diff
@@ -136,7 +186,69 @@ type notifier struct {
...
+	readyToMerge  *atomic.Uint64
+}
+
 type applyPollerBuilder struct {
+	tag             string
+	cfg             *Config
+	coprocessorHost *CoprocessorHost
+	engines         *Engines
+	sender          notifier
+	router          applyRouter
```
coocood: use pointer here because applyRouter embeds router which is a big struct.

tikv/raftstore/fsm_apply.go
```diff
@@ -136,7 +186,69 @@ type notifier struct {
...
+
+	enableSyncLog  bool
+	syncLogHint    bool
+	useDeleteRange bool
+}
+
+type waitSourceMergeState struct {
+	pendingEntries []*eraftpb.Entry
+	pendingMsgs    []Msg
+	readyToMerge   *atomic.Uint64
```
coocood: This library is convenient, we can use it to replace other atomic access values.
But it's a new library, we need to add an entry in `go.mod` file.

tikv/raftstore/fsm_apply.go
xiaobogaga: ok .

tikv/raftstore/fsm_apply.go
xiaobogaga: sorry, I would add this thanks.

xiaobogaga:
@coocood  PTAL thanks.

tikv/raftstore/fsm_apply.go
```diff
@@ -95,6 +109,42 @@ type execResult struct {
...
+
+/// applyResult has three kinds of types which are `None`, `ExecResultType` and `WaitMergeResourceType`.
+/// The data is different for `ExecResultType` and `WaitMergeResourceType`. For `ExecResultType`, the data
+/// is a ExecResult, and for `WaitMergeResourceType`, it is a `WaitMergeSource`.
+type applyResult struct {
+	tp   applyResultType
+	data interface{}
+}
+
+type waitMergeSource struct {
```
coocood: This struct can be removed, we can use `atomic.Uint64` directly.

tikv/raftstore/fsm_apply.go
```diff
@@ -136,7 +186,69 @@ type notifier struct {
...
+	appliedIndexTerm  uint64
+	metrics           ApplyMetrics
+}
+
+type applyDestroy struct {
+	regionId uint64
+}
+
+type catchUpLogs struct {
+	targetMailBox mailbox
```
coocood: `mailbox` should use pointer because it has shared state.

tikv/raftstore/fsm_apply.go
```diff
@@ -95,6 +109,42 @@ type execResult struct {
 	data interface{}
 }
 
+type applyResultType int
+
+const (
+	None applyResultType = iota
```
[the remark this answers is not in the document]
xiaobogaga: ok

xiaobogaga:
@coocood PTAL thanks.

PR #156, opened 2019-03-28
Port store fsm step 3

coocood:


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/fsm_store.go
```diff
@@ -698,58 +723,357 @@ func (bs *raftBatchSystem) shutDown() {
...
+		return false, errors.Errorf("tombstone peer [epoch: %s] received an invalid message %s, ignore it",
+			regionEpoch, msgType)
+	}
+	return false, nil
 }
 
 func (d *storeFsmDelegate) onRaftMessage(msg *rspb.RaftMessage) error {
-	return nil // TODO: stub
+	regionID := msg.RegionId
+	if err := d.ctx.router.send(regionID, Msg{Type: MsgTypeRaftMessage, Data: msg}); err != nil {
```
zhangjinpeng1987: How about channel is closed when shutdown?

tikv/raftstore/fsm_store.go
```diff
@@ -698,58 +723,357 @@ func (bs *raftBatchSystem) shutDown() {
...
+		// Target tombstone peer doesn't exist, so ignore it.
+		return nil
+	}
+	ok, err := d.checkMsg(msg)
+	if err != nil {
+		return err
+	}
+	if ok {
+		return nil
+	}
```
zhangjinpeng1987: Missing `maybe_create_peer`.

tikv/raftstore/fsm_store.go
```diff
@@ -698,58 +723,357 @@ func (bs *raftBatchSystem) shutDown() {
...
+		// Make sure the range of region from msg is covered by existing regions.
+		// If so, means that the region may be generated by some kinds of split
+		// and merge by catching logs. So there is no need to accept a snapshot.
+		if !isRangeCovered(meta, msg.StartKey, msg.EndKey) {
+			if maybeDestroySource(meta, regionID, existRegion.Id, msg.RegionEpoch) {
+				regionsToDestroy = append(regionsToDestroy, existRegion.Id)
+				continue
+			}
+		}
+		return false, nil
```
zhangjinpeng1987: When return from here, seems should not destroy regions?

tikv/raftstore/fsm_store.go
coocood: My mistake.

PR #157, opened 2019-03-28
Port splitCheckRun of worker, step 1

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/tikv/raftstore/coprocessor.go
+++ b/tikv/raftstore/coprocessor.go
@@ -21,16 +21,27 @@ type observerContext struct {
 	bypass bool
 }
 
+/// splitChecker is invoked during a split check scan, and decides to use
+/// which keys to split a region.
 type splitChecker interface {
+
+	/// onKv is a hook called for every kv scanned during split.
+	/// Return true to abort scan early.
 	onKv(_ *observerContext, _ splitCheckKeyEntry) bool
+
+	/// splitKeys returns the desired split keys.
 	splitKeys() [][]byte
+
+	/// approximateSplitKeys returns the split keys without scan.
 	approximateSplitKeys(_ *metapb.Region, _ *badger.DB) ([][]byte, error)
+
+	/// policy returns the policy.
 	policy() pdpb.CheckPolicy
 }
 
 type splitCheckerHost struct {
 	autoSplit bool
-	checkers  []*splitChecker
+	checkers  []splitChecker
 }
 
 func (spCheckerHost *splitCheckerHost) skip() bool {
@@ -38,7 +49,7 @@ func (spCheckerHost *splitCheckerHost) skip() bool {
 	return false
 }
 
-func (spCheckerHost *splitCheckerHost) onKv() bool {
+func (spCheckerHost *splitCheckerHost) onKv(observerCtx *observerContext, spCheKeyEntry splitCheckKeyEntry) bool {
 	// Todo: currently it is a place holder
 	return false
 }
@@ -48,7 +59,7 @@ func (spCheckerHost *splitCheckerHost) splitKeys() [][]byte {
 	return nil
 }
 
-func (spCheckerHost *splitCheckerHost) approximateSplitKeys(_ *metapb.Region, _ *badger.DB) ([][]byte, error) {
+func (spCheckerHost *splitCheckerHost) approximateSplitKeys(region *metapb.Region, db *badger.DB) ([][]byte, error) {
 	// Todo: currently it is a place holder
 	return nil, nil
 }
--- a/tikv/raftstore/coprocessor.go
+++ b/tikv/raftstore/coprocessor.go
@@ -1,7 +1,9 @@
 package raftstore
 
 import (
+	"github.com/coocood/badger"
 	"github.com/pingcap/kvproto/pkg/metapb"
+	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/kvproto/pkg/raft_cmdpb"
 	"github.com/zhangjinpeng1987/raft"
 )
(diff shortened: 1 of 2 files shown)


tikv/raftstore/worker.go
```diff
@@ -180,14 +182,80 @@ func newWorker(name string, wg *sync.WaitGroup) *worker {
...
 func (r *splitCheckRunner) run(t task) {
-	// TODO: stub
+	spCheckTask := t.data.(*splitCheckTask)
+	region := spCheckTask.region
+	regionId := region.Id
+	startKey := EncStartKey(region)
+	endKey := EncEndKey(region)
+	log.Debugf("executing task: [regionId: %d, startKey: %s, endKey: %s]", regionId,
+		hex.EncodeToString(startKey), hex.EncodeToString(endKey))
+	/// todo, a metrics method ignored here.
```
coocood: We can remove those metrics todos.

tikv/raftstore/worker.go
```diff
@@ -180,14 +182,80 @@ func newWorker(name string, wg *sync.WaitGroup) *worker {
...
+	host := r.coprocessorHost.newSplitCheckerHost(region, r.engine, spCheckTask.autoSplit,
+		spCheckTask.policy)
+	if host.skip() {
+		log.Debugf("skip split check, [regionId : %d]", regionId)
```
coocood: missing `return`.

tikv/raftstore/worker.go
```diff
@@ -180,14 +182,80 @@ func newWorker(name string, wg *sync.WaitGroup) *worker {
...
+		keys, err = host.approximateSplitKeys(region, r.engine)
+		if err == nil {
+			for i, k := range keys {
+				keys[i] = OriginKey(k)
+			}
+			break
+		}
+		log.Errorf("failed to get approximate split key, try scan way: [regionId: %d, err : %v]",
+			regionId, err)
+		fallthrough
```
coocood: Use `fallthrough` does avoid some duplicated code.
But we rarely use fallthough, it makes the logic more tangled, harder to read.

tikv/raftstore/worker.go
```diff
@@ -180,14 +182,80 @@ func newWorker(name string, wg *sync.WaitGroup) *worker {
...
+	var keys [][]byte
+	var err error
+	switch host.policy() {
+	case pdpb.CheckPolicy_APPROXIMATE:
+		keys, err = host.approximateSplitKeys(region, r.engine)
+		if err == nil {
+			for i, k := range keys {
+				keys[i] = OriginKey(k)
+			}
+			break
```
coocood: We rarely use `break` to break `switch` case, use `else` here is better.

tikv/raftstore/worker.go
```diff
@@ -180,14 +182,80 @@ func newWorker(name string, wg *sync.WaitGroup) *worker {
...
+			regionId, err)
+		fallthrough
+	case pdpb.CheckPolicy_SCAN:
+		keys, err = r.scanSplitKeys(host, region, startKey, endKey)
+		if err != nil {
+			log.Errorf("failed to scan split key: [regionId: %d, err: %v]", regionId, err)
+			return
+		}
+	}
+	if keys != nil && len(keys) != 0 {
```
coocood: It's safe to call `len(keys)` even if keys is nil.

tikv/raftstore/coprocessor.go
```diff
@@ -14,6 +16,48 @@ const (
...
+type splitChecker interface {
+	onKv(_ *observerContext, _ splitCheckKeyEntry) bool
+	splitKeys() [][]byte
+	approximateSplitKeys(_ *metapb.Region, _ *badger.DB) ([][]byte, error)
+	policy() pdpb.CheckPolicy
+}
+
+type splitCheckerHost struct {
+	autoSplit bool
+	checkers  []*splitChecker
```
coocood: Never use pointer for interface types.

tikv/raftstore/coprocessor.go
```diff
@@ -14,6 +16,48 @@ const (
...
+}
+
+func (spCheckerHost *splitCheckerHost) skip() bool {
+	//Todo, currently it is a place holder
+	return false
+}
+
+func (spCheckerHost *splitCheckerHost) onKv() bool {
```
coocood: missing arguments.

tikv/raftstore/coprocessor.go
```diff
@@ -14,6 +16,48 @@ const (
 	RegionChangeEvent_Destroy
 )
 
+type observerContext struct {
+	region *metapb.Region
+	bypass bool
+}
+
+type splitChecker interface {
```
coocood: We also need to port comments.

xiaobogaga: ok

tikv/raftstore/coprocessor.go
xiaobogaga: sorry

tikv/raftstore/coprocessor.go
xiaobogaga: yes

tikv/raftstore/worker.go
xiaobogaga: yes

tikv/raftstore/worker.go
xiaobogaga: sorry, I would fix this

tikv/raftstore/worker.go
xiaobogaga: ok, thanks

coocood:
LGTM

PR #159, opened 2019-03-31
port apply fsm step 3

coocood:
Also replaced pb type `RaftLocalState` and `RaftApplyState` to simply struct type to reduce clone cost.

### DIFF
--- a/tikv/raftstore/fsm_apply.go
+++ b/tikv/raftstore/fsm_apply.go
@@ -1,12 +1,20 @@
 package raftstore
 
 import (
+	"fmt"
+	"github.com/coocood/badger/y"
+	"github.com/ngaut/log"
 	"github.com/pingcap/kvproto/pkg/eraftpb"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/raft_cmdpb"
 	rspb "github.com/pingcap/kvproto/pkg/raft_serverpb"
 	"github.com/uber-go/atomic"
 	"time"
+)
+
+const (
+	WriteBatchMaxKeys  = 128
+	DefaultApplyWBSize = 4 * 1024
 )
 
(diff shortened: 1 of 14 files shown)


tikv/raftstore/fsm_apply.go
```diff
@@ -446,23 +615,121 @@ func (d *applyDelegate) updateMetrics(aCtx *applyContext) {
...
+	wb.Set(ApplyStateKey(d.region.Id), d.applyState.Marshal())
 }
 
 func (d *applyDelegate) handleRaftEntryNormal(aCtx *applyContext, entry *eraftpb.Entry) applyResult {
-	return applyResult{} // TODO: stub
+	index := entry.Index
+	term := entry.Term
+	if len(entry.Data) > 0 {
+		cmd := new(raft_cmdpb.RaftCmdRequest)
+		err := cmd.Unmarshal(entry.Data)
```
zhangjinpeng1987: Should check the term and index after unmarshal from bytes.

tikv/raftstore/fsm_apply.go
```diff
@@ -446,23 +615,121 @@ func (d *applyDelegate) updateMetrics(aCtx *applyContext) {
...
+		panic("unreachable")
+	}
 }
 
 func (d *applyDelegate) findCallback(index, term uint64, isConfChange bool) Callback {
-	return nil // TODO: stub
+	regionID := d.region.Id
+	peerID := d.id
+	if isConfChange {
+		for {
```
zhangjinpeng1987: Seems this loop is no need.

tikv/raftstore/fsm_apply.go
coocood: but tikv doesn't check term and index.

zhangjinpeng1987: in parse_data_at.

PR #160, opened 2019-04-01
Port splitCheckRun of worker, step 2

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/tikv/raftstore/coprocessor.go
+++ b/tikv/raftstore/coprocessor.go
@@ -21,16 +21,27 @@ type observerContext struct {
 	bypass bool
 }
 
+/// splitChecker is invoked during a split check scan, and decides to use
+/// which keys to split a region.
 type splitChecker interface {
+
+	/// onKv is a hook called for every kv scanned during split.
+	/// Return true to abort scan early.
 	onKv(_ *observerContext, _ splitCheckKeyEntry) bool
+
+	/// splitKeys returns the desired split keys.
 	splitKeys() [][]byte
+
+	/// approximateSplitKeys returns the split keys without scan.
 	approximateSplitKeys(_ *metapb.Region, _ *badger.DB) ([][]byte, error)
+
+	/// policy returns the policy.
 	policy() pdpb.CheckPolicy
 }
 
 type splitCheckerHost struct {
 	autoSplit bool
-	checkers  []*splitChecker
+	checkers  []splitChecker
 }
 
 func (spCheckerHost *splitCheckerHost) skip() bool {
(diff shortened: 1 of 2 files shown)


tikv/raftstore/worker.go
```diff
@@ -244,11 +245,30 @@ func (r *splitCheckRunner) run(t task) {
...
+	ite := reader.GetIter()
+	defer reader.Close()
+	for ite.Seek(startKey); ite.Valid(); ite.Next() {
+		item := ite.Item()
+		key := item.KeyCopy(nil)
+		if exceedEndKey(key, endKey) {
+			break
+		}
+		if value, err := item.Value(); err == nil {
+			spCheckerHost.onKv(region, &splitCheckKeyEntry{key: key, valueSize: uint64(len(value)),})
```
coocood: Should check the returned bool value?

xiaobogaga: yes. I would fix.

tikv/raftstore/worker.go
```diff
@@ -244,11 +245,30 @@ func (r *splitCheckRunner) run(t task) {
...
 	startKey []byte, endKey []byte) ([][]byte, error) {
-	/// Todo, currently it is a place holder
-	return nil, nil
+	txn := r.engine.NewTransaction(false)
+	reader := dbreader.NewDBReader(startKey, endKey, txn, 0)
+	ite := reader.GetIter()
+	defer reader.Close()
+	for ite.Seek(startKey); ite.Valid(); ite.Next() {
+		item := ite.Item()
+		key := item.KeyCopy(nil)
```
coocood: copy every key during scanSplitKeys is very expensive.
Only a few keys needed to copy.
We can avoid copy here and make sure the key is copied in split checkers

xiaobogaga: ok, thanks.

tikv/raftstore/coprocessor.go
```diff
@@ -27,7 +39,7 @@ type splitChecker interface {
 
 	/// onKv is a hook called for every kv scanned during split.
 	/// Return true to abort scan early.
-	onKv(_ *observerContext, _ splitCheckKeyEntry) bool
+	onKv(_ *observerContext, _ *splitCheckKeyEntry) bool
```
coocood: We don't need to change splitCheckKeyEntry to pointer here.

tikv/raftstore/coprocessor.go
```diff
@@ -45,31 +57,51 @@ type splitCheckerHost struct {
...
-}
-
-func (spCheckerHost *splitCheckerHost) onKv(observerCtx *observerContext, spCheKeyEntry splitCheckKeyEntry) bool {
-	// Todo: currently it is a place holder
+	return len(spCheckerHost.checkers) == 0
+}
+
+/// onKv is a hook called for every check during split.
+/// Return true means abort early.
+func (spCheckerHost *splitCheckerHost) onKv(region *metapb.Region, spCheKeyEntry *splitCheckKeyEntry) bool {
```
coocood: use struct for `splitCheckKeyEntry` can avoid one memory allocation

xiaobogaga:
@coocood PTAL again, thanks.

coocood:
LGTM

PR #161, opened 2019-04-01
add more pd client methods

disksing:

### DIFF
--- a/pd/client.go
+++ b/pd/client.go
@@ -273,7 +273,6 @@ func (c *client) IsBootstrapped(ctx context.Context) (bool, error) {
 		return false, err
 	}
 	if herr := resp.Header.GetError(); herr != nil {
-		log.Info(herr)
 		return false, errors.New(herr.String())
 	}
 	return resp.Bootstrapped, nil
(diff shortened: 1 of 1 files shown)


pd/client.go
```diff
@@ -265,13 +289,165 @@ func (c *client) PutStore(ctx context.Context, store *metapb.Store) error {
 	if err != nil {
 		return err
 	}
-	if resp.Header.GetError() != nil {
-		log.Info(resp.Header.GetError())
-		return nil
+	if herr := resp.Header.GetError(); herr != nil {
+		log.Info(herr)
```
coocood: Since we returned the header error, maybe we don't need to log it?

pd/client.go
```diff
@@ -265,13 +289,165 @@ func (c *client) PutStore(ctx context.Context, store *metapb.Store) error {
...
+	resp, err := c.pdClient().GetStore(ctx, &pdpb.GetStoreRequest{
+		Header:  c.requestHeader(),
+		StoreId: storeID,
+	})
+	cancel()
+	if err != nil {
+		return nil, err
+	}
+	if herr := resp.Header.GetError(); herr != nil {
+		log.Info(herr)
```
coocood: ditto

PR #164, opened 2019-04-02
Port splitCheck and SplitCheckObserver implementations, step 2

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,6 +1,8 @@
 module github.com/ngaut/unistore
 
 require (
+	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
+	github.com/alecthomas/units v0.0.0-20151022065526-2efee857e7cf // indirect
 	github.com/coocood/badger v1.5.1-0.20181115111105-250ee6037b80
 	github.com/cznic/mathutil v0.0.0-20181021201202-eba54fb065b7
 	github.com/dgryski/go-farm v0.0.0-20190104051053-3adb47b1fb0f
@@ -24,6 +26,7 @@ require (
 	golang.org/x/time v0.0.0-20181108054448-85acf8d2951c
 	google.golang.org/genproto v0.0.0-20181016170114-94acd270e44e // indirect
 	google.golang.org/grpc v1.16.0
+	gopkg.in/alecthomas/kingpin.v2 v2.2.6 // indirect
 	gopkg.in/mgo.v2 v2.0.0-20180705113604-9856a29383ce // indirect
 )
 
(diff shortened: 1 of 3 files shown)


tikv/raftstore/coprocessor.go
```diff
@@ -5,6 +5,7 @@ import (
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/pdpb"
 	"github.com/pingcap/kvproto/pkg/raft_cmdpb"
+	"github.com/prometheus/common/log"
```
coocood: imported the wrong package.

go.mod
```diff
@@ -1,6 +1,8 @@
 module github.com/ngaut/unistore
 
 require (
+	github.com/alecthomas/template v0.0.0-20160405071501-a0175ee3bccc // indirect
```
coocood: introduced by wrong log package "github.com/prometheus/common/log"

tikv/raftstore/worker.go
```diff
@@ -270,6 +273,8 @@ func (r *splitCheckRunner) scanSplitKeys(spCheckerHost *splitCheckerHost, region
 			return nil, errors.Trace(err)
 		}
 	}
+	// Close the reader early, and in the `onKv` method above, some keys must be copied.
+	reader.Close()
```
coocood: The reader doesn't hold any lock, we don't need to close it early.

tikv/raftstore/coprocessor.go
```diff
@@ -68,21 +69,56 @@ func newSizeSplitChecker(maxSize, splitSize, batchSplitLimit uint64, policy pdpb
...
 
 func (checker *sizeSplitChecker) getSplitKeys() [][]byte {
-	//Todo, currently it is a place holder
+	// Make sure not to split when less than maxSize for last part
+	if checker.currentSize+checker.splitSize < checker.maxSize {
+		if len(checker.splitKeys) != 0 {
+			checker.splitKeys = checker.splitKeys[:len(checker.splitKeys)-1]
+		}
+	}
+	if len(checker.splitKeys) != 0 {
```
coocood: We can safely remove this check.

tikv/raftstore/coprocessor.go

tikv/raftstore/worker.go
xiaobogaga: yes, thanks.

tikv/raftstore/coprocessor.go
xiaobogaga: yes, thanks.

xiaobogaga:
@coocood PTAL again, thanks.

coocood:
LGTM

PR #167, opened 2019-04-03
port apply fsm step 4

coocood:
Because unistore doesn't have CF, executing write command is complex, the code is not directly ported, need careful review.

### DIFF
--- a/tikv/raftstore/fsm_apply.go
+++ b/tikv/raftstore/fsm_apply.go
@@ -4,6 +4,8 @@ import (
 	"fmt"
 	"github.com/coocood/badger/y"
 	"github.com/ngaut/log"
+	"github.com/ngaut/unistore/tikv/mvcc"
+	"github.com/pingcap/errors"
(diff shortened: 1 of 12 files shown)


tikv/raftstore/fsm_apply.go
```diff
@@ -833,26 +840,215 @@ func (d *applyDelegate) execRaftCmd(aCtx *applyContext, req *raft_cmdpb.RaftCmdR
...
+
+type writeCmdOps struct {
+	prewrites []prewriteOp
+	commits   []commitOp
+	rollbacks []rollbackOp
+	delRanges []*raft_cmdpb.DeleteRangeRequest
+}
+
+// createWriteCmdOps regroups requests into operations.
+func (d *applyDelegate) createWriteCmdOps(requests []*raft_cmdpb.Request) (ops writeCmdOps) {
```
zhangjinpeng1987: Better to add a test to cover this function.

coocood: I'll do it in future PR.

PR #168, opened 2019-04-04
Port splitCheck and splitCheckObserver implementations, step 3

xiaobogaga:
@coocood PTAL thanks

### DIFF
--- a/tikv/raftstore/coprocessor.go
+++ b/tikv/raftstore/coprocessor.go
@@ -325,7 +325,6 @@ func (observer *keysSplitCheckObserver) addChecker(obCtx *observerContext, host 
 }
 
 type tableSplitChecker struct {
-	firstEncodedTablePrefix []byte
 	splitKey                []byte
 	checkPolicy             pdpb.CheckPolicy
 }
@@ -334,9 +333,8 @@ func newTableSplitCheckerByPolicy(policy pdpb.CheckPolicy) *tableSplitChecker {
 	return &tableSplitChecker{checkPolicy: policy}
 }
 
-func newTableSplitChecker(firstEncodedTablePrefix, splitKey []byte, policy pdpb.CheckPolicy) *tableSplitChecker {
+func newTableSplitChecker(splitKey []byte, policy pdpb.CheckPolicy) *tableSplitChecker {
 	return &tableSplitChecker{
-		firstEncodedTablePrefix: firstEncodedTablePrefix,
 		splitKey:                splitKey,
 		checkPolicy:             policy,
 	}
@@ -361,8 +359,7 @@ func (checker *tableSplitChecker) onKv(obCtx *observerContext, spCheckKeyEntry s
 	}
 	currentEncodedKey := OriginKey(spCheckKeyEntry.key)
 	var splitKey []byte
-	// No need to check firstEncodedTablePrefix, since it is always nil.
-	if len(checker.firstEncodedTablePrefix) == 0 && isTableKey(currentEncodedKey) {
+	if isTableKey(currentEncodedKey) {
 		splitKey = currentEncodedKey
 	}
 	if len(splitKey) != 0 {
@@ -427,7 +424,7 @@ func (observer *tableSplitCheckObserver) addChecker(obCtx *observerContext, host
 		// Note that table id does not grow by 1, so have to use encodedEndKey to extract a table prefix.
 		// See more: https://github.com/pingcap/tidb/issues/4727.
 		splitKey := extractTablePrefix(encodedEndKey)
-		host.addChecker(newTableSplitChecker(nil, splitKey, policy))
+		host.addChecker(newTableSplitChecker(splitKey, policy))
 	}
 }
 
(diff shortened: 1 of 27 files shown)


tikv/raftstore/coprocessor.go
```diff
@@ -1,11 +1,17 @@
 package raftstore
 
 import (
+	"bytes"
+	"fmt"
 	"github.com/coocood/badger"
 	"github.com/ngaut/log"
+	"github.com/ngaut/unistore/tikv"
```
coocood: importing this package will cause dependency loop in the future, we need to move the sharding key function to `mvcc` package.

tikv/raftstore/coprocessor.go
```diff
@@ -97,8 +103,11 @@ func (checker *sizeSplitChecker) onKv(obCtx *observerContext, spCheckKeyEntry sp
 func (checker *sizeSplitChecker) getSplitKeys() [][]byte {
 	// Make sure not to split when less than maxSize for last part
 	if checker.currentSize+checker.splitSize < checker.maxSize {
-		if len(checker.splitKeys) != 0 {
-			checker.splitKeys = checker.splitKeys[:len(checker.splitKeys)-1]
+		splitKeyLen := len(checker.splitKeys)
+		if splitKeyLen != 0 {
+			// Help gc
```
coocood: No need to do this because the checker is a short lived object.

tikv/raftstore/coprocessor.go
```diff
@@ -203,6 +212,8 @@ func (checker *halfSplitChecker) getSplitKeys() [][]byte {
 	}
 	dataKey := checker.buckets[mid]
 	checker.buckets[mid] = checker.buckets[bucketLen-1]
+	// Help gc
+	checker.buckets[bucketLen-1] = nil
```
coocood: ditto

tikv/raftstore/coprocessor.go
xiaobogaga: yes, I would fix this.

tikv/raftstore/coprocessor.go
xiaobogaga: ok, thanks.

tikv/raftstore/coprocessor.go

tikv/raftstore/coprocessor.go
```diff
@@ -334,14 +349,57 @@ func newTableSplitChecker(firstEncodedTablePrefix, splitKey []byte, policy pdpb.
 	}
 }
 
+func isTableKey(encodedKey []byte) bool {
+	if tikv.IsShardingEnabled() {
+		return len(encodedKey) >= tablecodec.TableSplitKeyLen &&
```
coocood: the `TableSplitKeyLen` is wrong when sharding enabled.
Should be 2 bytes longer than not enabled.
We can define it in this package for now.

xiaobogaga: got it.

tikv/raftstore/coprocessor.go
```diff
@@ -334,14 +349,57 @@ func newTableSplitChecker(firstEncodedTablePrefix, splitKey []byte, policy pdpb.
...
+// of currentEncodedKey
 func (checker *tableSplitChecker) onKv(obCtx *observerContext, spCheckKeyEntry splitCheckKeyEntry) bool {
-	//Todo, currently it is a place holder
+	if len(checker.splitKey) != 0 {
+		return true
+	}
+	currentEncodedKey := OriginKey(spCheckKeyEntry.key)
+	var splitKey []byte
+	// Different tables or meet the very first table key of this region.
+	if !isSameTable(checker.firstEncodedTablePrefix, currentEncodedKey) || isTableKey(currentEncodedKey) {
```
coocood: The logic is different than tikv.
Only when `firstEncodedTablePrefix` is empty, `isTableKey` sets `splitKey`

xiaobogaga:
@coocood PTAL again, thanks

tikv/raftstore/coprocessor.go
```diff
@@ -341,51 +335,46 @@ func newTableSplitCheckerByPolicy(policy pdpb.CheckPolicy) *tableSplitChecker {
...
-// of currentEncodedKey
 func (checker *tableSplitChecker) onKv(obCtx *observerContext, spCheckKeyEntry splitCheckKeyEntry) bool {
 	if len(checker.splitKey) != 0 {
 		return true
 	}
 	currentEncodedKey := OriginKey(spCheckKeyEntry.key)
 	var splitKey []byte
-	// Different tables or meet the very first table key of this region.
-	if !isSameTable(checker.firstEncodedTablePrefix, currentEncodedKey) || isTableKey(currentEncodedKey) {
+	// No need to check firstEncodedTablePrefix, since it is always nil.
```
coocood: There is no `firstEncodedTablePrefix`

xiaobogaga:
@coocood PTAL again, thanks

coocood:
LGTM

PR #170, opened 2019-04-05
port apply fsm step 5

coocood:


coocood:
@zhangjinpeng1987 PTAL

tikv/raftstore/util.go
```diff
@@ -456,3 +468,11 @@ func checkPeerID(req *raft_cmdpb.RaftCmdRequest, peerID uint64) error {
 	}
 	return errors.Errorf("mismatch peer id %d != %d", peer.Id, peerID)
 }
+
+func CloneMsg(origin, cloned proto.Message) error {
```
zhangjinpeng1987: I think cloned should be a pointer type?

coocood: It's an interface, the interface contains a pointer.

zhangjinpeng1987: got it

tikv/raftstore/fsm_apply.go
```diff
@@ -1037,34 +1053,300 @@ func (d *applyDelegate) createWriteCmdOps(requests []*raft_cmdpb.Request) (ops w
...
+		if p := findPeer(region, storeID); p != nil {
+			exist = true
+			if !p.IsLearner || p.Id != peer.Id {
+				errMsg := fmt.Sprintf("%s can't add duplicated peer, peer %s, region %s",
+					d.tag, p, d.region)
+				log.Error(errMsg)
+				err = errors.New(errMsg)
+				return
+			} else {
+				p.IsLearner = true
```
zhangjinpeng1987: Should be false.

tikv/raftstore/fsm_apply.go
```diff
@@ -1037,34 +1053,300 @@ func (d *applyDelegate) createWriteCmdOps(requests []*raft_cmdpb.Request) (ops w
...
+	}
+	log.Infof("%s split region %s, keys %v", d.tag, d.region, keys)
+	derived.RegionEpoch.Version += uint64(newRegionCnt)
+	// Note that the split requests only contain ids for new regions, so we need
+	// to handle new regions and old region separately.
+	if !rightDerive {
+		derived.EndKey = keys[1]
+		keys = keys[1:]
+		regions = append(regions, derived)
+	}
```
zhangjinpeng1987: Else you should drop the last key in keys?

coocood: No need, the last key will not be used.

PR #173, opened 2019-04-09
tikv: implement pessimistic lock protocol

tiancaiamao:
PTAL @coocood

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -407,6 +407,7 @@ const (
 
 func (store *MVCCStore) Rollback(reqCtx *requestCtx, keys [][]byte, startTS uint64) error {
 	store.updateLatestTS(startTS)
+	keys = deduplicateKeys(keys)
 	hashVals := keysToHashVals(keys...)
 	regCtx := reqCtx.regCtx
 	lockBatch := newWriteLockBatch(reqCtx)
@@ -429,6 +430,18 @@ func (store *MVCCStore) Rollback(reqCtx *requestCtx, keys [][]byte, startTS uint
 	}
 	err := store.writeLocks(lockBatch)
 	return errors.Trace(err)
+}
+
+func deduplicateKeys(keys [][]byte) [][]byte {
+	m := make(map[string]struct{})
+	deduped := make([][]byte, 0, len(keys))
+	for _, key := range keys {
+		if _, ok := m[string(key)]; !ok {
+			m[string(key)] = struct{}{}
+			deduped = append(deduped, key)
+		}
+	}
+	return deduped
 }
 
 func (store *MVCCStore) rollbackKeyReadLock(batch *writeLockBatch, key []byte, startTS uint64) (status int) {
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -8,6 +8,7 @@ import (
 	"sync/atomic"
 
 	"github.com/coocood/badger"
+	"github.com/coocood/badger/y"
 	"github.com/cznic/mathutil"
 	"github.com/juju/errors"
 	"github.com/ngaut/log"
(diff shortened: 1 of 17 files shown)


coocood:
PLock should not block read, but seems there is no logic for it.

coocood:
Prewrite should detect PLock written by self, and update it to prewrite lock.

tikv/mvcc.go
```diff
@@ -144,6 +152,84 @@ func (store *MVCCStore) Prewrite(reqCtx *requestCtx, mutations []*kvrpcpb.Mutati
...
+	for _, m := range mutations {
+		lock, err := store.checkConflictInLockStore(reqCtx, m, startTS)
+		if err != nil {
+			anyError = true
+		}
+		if lock != nil {
+			if lock.Op == uint8(kvrpcpb.Op_PessimisticLock) {
+				// lockstore doesn't support update for now, delete the lock first.
+				lockBatch.delete(m.Key)
+				hasPessimistic = true
```
lysu: maybe we should only set `hasPessimistic` flag just for in current `m`, other mutators also need check conflict?

coocood: pessimistic lock never conflicts in prewrite.

tikv/mvcc.go
```diff
@@ -130,11 +136,13 @@ func (store *MVCCStore) Prewrite(reqCtx *requestCtx, mutations []*kvrpcpb.Mutati
...
 	for _, m := range mutations {
-		duplicate, err := store.checkPrewriteInLockStore(reqCtx, m, startTS)
+		lock, err := store.checkConflictInLockStore(reqCtx, m, startTS)
 		if err != nil {
 			anyError = true
 		}
-		if duplicate {
+		if lock != nil {
+			// The lock already exists, it's a duplicate command, we can return directly.
+			y.Assert(lock.Op == uint8(kvrpcpb.Op_PessimisticLock))
```
tiancaiamao: This could be done in TiDB too, and the performance should be better.

tikv/mvcc.go
```diff
@@ -229,14 +324,14 @@ func (store *MVCCStore) checkPrewriteInLockStore(
...
-func (store *MVCCStore) checkPrewriteInDB(
+func (store *MVCCStore) checkConflictInDB(
 	req *requestCtx, txn *badger.Txn, item *badger.Item, startTS uint64) (userMeta mvcc.DBUserMeta, val []byte, err error) {
 	if item == nil {
 		return nil, nil, nil
 	}
 	userMeta = mvcc.DBUserMeta(item.UserMeta())
 	if userMeta.CommitTS() > startTS {
-		return nil, nil, ErrRetryable("write conflict")
+		return nil, nil, ErrRetryable("[write conflict]")
```
tiancaiamao: This kind of code is fragile IMHO

tikv/mvcc.go
```diff
@@ -144,6 +152,84 @@ func (store *MVCCStore) Prewrite(reqCtx *requestCtx, mutations []*kvrpcpb.Mutati
...
+			anyError = true
+		}
+		if lock != nil {
+			if lock.Op == uint8(kvrpcpb.Op_PessimisticLock) {
+				// lockstore doesn't support update for now, delete the lock first.
+				lockBatch.delete(m.Key)
+				hasPessimistic = true
+			} else {
+				// duplicated command.
+				return nil
```
tiancaiamao: Should it be continue here?

tikv/mvcc.go
tiancaiamao: Done

tikv/mvcc.go
```diff
@@ -119,22 +129,108 @@ func (store *MVCCStore) updateLatestTS(ts uint64) {
...
 	// Must check the LockStore first.
-	for _, m := range mutations {
-		duplicate, err := store.checkPrewriteInLockStore(reqCtx, m, startTS)
+	for i, m := range mutations {
+		lock, err := store.checkConflictInLockStore(reqCtx, m, req.StartVersion)
 		if err != nil {
 			anyError = true
 		}
-		if duplicate {
+		if hasPessimistic && req.PessimisticLockExist[i] {
```
youjiali1995: What's the use of req.PessimisticLockExist[i]? Only all pessimistic locks exist, prewrite can succeed.

coocood: non-unique index and delete index don't need to be locked.

util/lockwaiter/lockwaiter.go
```diff
@@ -0,0 +1,126 @@
...
+	Position Position
+	CommitTS uint64
+}
+
+const WaitTimeout Position = -1
+
+func (w *Waiter) Wait() Result {
+	select {
+	case <-time.After(w.timeout):
+		return Result{Position: WaitTimeout}
```
youjiali1995: When timeout, the waiter should clean up itself.

coocood:
replaced by https://github.com/ngaut/unistore/pull/202

PR #174, opened 2019-04-09
introduce DBWriter interface

coocood:
So we can optionally enable raft.

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -1,14 +1,17 @@
 package tikv
 
 import (
+	"bufio"
 	"bytes"
 	"encoding/binary"
+	"io"
 	"math"
+	"os"
 	"sync"
 	"sync/atomic"
+	"unsafe"
 
 	"github.com/coocood/badger"
-	"github.com/cznic/mathutil"
 	"github.com/juju/errors"
 	"github.com/ngaut/log"
 	"github.com/ngaut/unistore/lockstore"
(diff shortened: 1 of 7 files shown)


coocood:
@zhangjinpeng1987 PTAL

tikv/mvcc.go
```diff
@@ -108,15 +125,40 @@ func (store *MVCCStore) Close() error {
...
+	it := store.lockStore.NewIterator()
+	hdrBuf := make([]byte, 8)
+	hdr := (*lockEntryHdr)(unsafe.Pointer(&hdrBuf[0]))
+	for it.SeekToFirst(); it.Valid(); it.Next() {
+		hdr.keyLen = uint32(len(it.Key()))
+		hdr.valLen = uint32(len(it.Value()))
+		writer.Write(hdrBuf)
+		writer.Write(it.Key())
+		writer.Write(it.Value())
+		cnt++
```
zhangjinpeng1987: Seems cnt not used?

coocood: I didn't modify the code, just move around, maybe later we can use it to debug.

tikv/write.go
```diff
@@ -214,114 +177,212 @@ func (w *writeLockWorker) run() {
...
+
+func (writer *dbWriter) Close() {
+	close(writer.closeCh)
+	writer.wg.Wait()
+}
+
+func (writer *dbWriter) Write(batch mvcc.WriteBatch) error {
+	wb := batch.(*writeBatch)
+	if len(wb.dbBatch.entries) > 0 {
+		// We must delete lock after commit succeed, or there will be inconsistency.
```
zhangjinpeng1987: I think this line comment can be moved to L275?

coocood: ok

tikv/write.go
```diff
@@ -214,114 +177,212 @@ func (w *writeLockWorker) run() {
...
+		atomic.CompareAndSwapUint64(&writer.latestTS, latestTS, ts)
+	}
+}
+
+const delRangeBatchSize = 4096
+
+func (writer *dbWriter) DeleteRange(startKey, endKey []byte, latchHandle mvcc.LatchHandle) error {
+	keys := make([][]byte, 0, delRangeBatchSize)
+	oldStartKey := mvcc.EncodeOldKey(startKey, maxSystemTS)
+	oldEndKey := mvcc.EncodeOldKey(endKey, maxSystemTS)
```
zhangjinpeng1987: If the range is [start, end), i think you don't need to calculate oldEndKey, just use endkey for old cf.

tikv/write.go
```diff
@@ -214,114 +177,212 @@ func (w *writeLockWorker) run() {
...
-		if err != nil {
-			return errors.Trace(err)
-		}
-		if cap(keyBuf) < int(hdr.keyLen) {
-			keyBuf = make([]byte, hdr.keyLen)
-		}
-		if cap(valBuf) < int(hdr.valLen) {
-			valBuf = make([]byte, hdr.valLen)
+		keys = append(keys, key)
+		if len(keys) == delRangeBatchSize {
```
zhangjinpeng1987: So the keys just part of the result? Should involve multiple `collectRangeKyes` and `deleteKeysInBatch` at L346?

tikv/write.go
coocood: But `endKey` has prefix 't', `oldStartKey` and `oldEndKey` has prefix `u`.

PR #177, opened 2019-04-10
Port regionWorker, step 3

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/lockstore/lockstore.go
+++ b/lockstore/lockstore.go
@@ -19,8 +19,8 @@ type MemStore struct {
 	arenaPtr unsafe.Pointer
 
 	// We only consume 2 bits for a random height call.
-	rand rand.Source64
-	size int
+	rand   rand.Source64
+	length int
 }
 
 const (
@@ -273,7 +273,7 @@ func (ls *MemStore) Insert(key []byte, v []byte) bool {
 		}
 		prev[i].setNextAddr(i, x.addr)
 	}
-	ls.size += 1
+	ls.length += 1
 	return true
 }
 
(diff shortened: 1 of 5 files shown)


xiaobogaga:
@coocood  PTAL, thanks.

tikv/raftstore/worker.go
```diff
@@ -290,7 +291,75 @@ type pendingDeleteRanges struct {
...
+
+func (peerInfo *stalePeerInfo) serializeStalePeerInfo() []byte {
+	if timeoutData, err := peerInfo.timeout.MarshalBinary(); err != nil {
+		log.Errorf("serialize stale Peer info failed, [regionId: %d, endKey: %v]", peerInfo.regionId, peerInfo.endKey)
+		return nil
+	} else {
+		buf := make([]byte, 8+4+4)
+		binary.LittleEndian.PutUint64(buf, peerInfo.regionId)
+		binary.LittleEndian.PutUint32(buf[8:], uint32(len(peerInfo.endKey)))
+		binary.LittleEndian.PutUint32(buf[12:], uint32(len(timeoutData)))
```
coocood: We can just store a single int64 for time.
```
        t := time.Now()
	n := t.UnixNano()
	t = time.Unix(0, n)
```
And we also don't need to store the length of endkey, if the total length of value is known, and other two fields takes 16 bytes.

lockstore/lockstore.go
```diff
@@ -20,6 +20,7 @@ type MemStore struct {
 
 	// We only consume 2 bits for a random height call.
 	rand rand.Source64
+	size int
```
coocood: Use `length` is better.

tikv/raftstore/worker.go
```diff
@@ -290,7 +291,75 @@ type pendingDeleteRanges struct {
...
+				startKey: startKey,
+				endKey:   peerInfo.endKey,
+				regionId: peerInfo.regionId,
+			})
+		}
+	}
+	return
+}
+
+type stalePeerInfo struct {
```
coocood: You can define this type like this
```
type stalePeerInfo struct {
    data []byte
}
```
Then define methods to get those fields.
So we don't need to call serialize and deserialize.

xiaobogaga:
@coocood  PTAL again, thanks.

coocood:
LGTM

PR #178, opened 2019-04-12
Port raftLogGcWorker and its test.

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/tikv/raftstore/worker.go
+++ b/tikv/raftstore/worker.go
@@ -394,7 +394,7 @@ func (snapCtx *snapContext) insertPendingDeleteRange(regionId uint64, startKey, 
 	}
 	snapCtx.cleanUpOverlapRanges(startKey, endKey)
 	log.Infof("register deleting data in range. [regionId: %d, startKey: %s, endKey: %s]", regionId,
-		regionId, hex.EncodeToString(startKey), hex.EncodeToString(endKey))
+		hex.EncodeToString(startKey), hex.EncodeToString(endKey))
 	timeout := time.Now().Add(snapCtx.cleanStalePeerDelay)
 	snapCtx.pendingDeleteRanges.insert(regionId, startKey, endKey, timeout)
 	return true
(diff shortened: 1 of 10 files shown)


xiaobogaga:
@coocood PTAL thanks.

tikv/raftstore/worker.go
```diff
@@ -624,11 +624,88 @@ func (r *regionRunner) shutdown() {
...
+	firstIdx := startIdx
+	if firstIdx == 0 {
+		firstIdx = endIdx
+		err := raftDb.View(func(txn *badger.Txn) error {
+			startKey := RaftLogKey(regionId, 0)
+			ite := txn.NewIterator(badger.DefaultIteratorOptions)
+			defer ite.Close()
+			if ite.Seek(startKey); ite.Valid() {
+				var err error
+				if firstIdx, err = RaftLogIndex(ite.Item().KeyCopy(nil)); err != nil {
```
coocood: We don't need to copy the key because the key is not referenced outside.

tikv/raftstore/worker.go
```diff
@@ -624,11 +624,88 @@ func (r *regionRunner) shutdown() {
...
+			return 0, err
+		}
+	}
+
+	if firstIdx >= endIdx {
+		log.Infof("no need to gc, [regionId: %d]", regionId)
+		return 0, nil
+	}
+
+	counter := uint64(0)
```
coocood: Why use `counter` instead of `endIdx - firstIdx`?

xiaobogaga: because error might happen and a partial raft log are gced during `writeToRaft`. I think it might be useful collecting raft log gc size even when error occurs, so by using counter, it can do this. For tikv, it just return `0` when error occurs.

coocood: No, the transaction is atomic, either all succeed or all fail.

xiaobogaga:
@coocood  PTAL again, thanks.

coocood:
LGTM

PR #179, opened 2019-04-16
Port splitCheck and SplitCheckObserver test.

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/tikv/raftstore/coprocessor_test.go
+++ b/tikv/raftstore/coprocessor_test.go
@@ -40,7 +40,6 @@ func newSplitCheckFsm(cap int) (chan Msg, fsm) {
 }
 
 func newTestRouter() (*router, chan Msg) {
-
 	ctrlSender, ctrlFsm := newSplitCheckFsm(10)
 	normarlSender, normalFsm := newSplitCheckFsm(10)
 	ctrlBox := newMailbox(ctrlSender, ctrlFsm)
@@ -220,16 +219,6 @@ func TestTableSplitCheckObserver(t *testing.T) {
 	defer cleanUpTestEngineData(engines)
 
 	region := newTestSplitCheckRegion()
-
-	config := newDefaultSplitCheckConfig()
-	// Enable table split.
-	config.splitRegionOnTable = true
-	// Try to "disable" size split.
-	config.regionMaxSize = uint64(2 * 1024 * 1024 * 1024)
-	config.regionSplitSize = uint64(1024 * 1024 * 1024)
-	// Try to "disable" keys split
-	config.regionMaxKeys = uint64(2000000000)
-	config.regionSplitKeys = uint64(1000000000)
 
 	// only create table split observer and ignore approximate.
 	router, receiver := newTestRouter()
--- a/tikv/raftstore/coprocessor_test.go
+++ b/tikv/raftstore/coprocessor_test.go
@@ -64,8 +64,8 @@ func newTestSplitCheckRegion() *metapb.Region {
 			ConfVer: 5,
 			Version: 2,
 		},
-		StartKey: codec.EncodeInt(tablecodec.TablePrefix(), int64(0)),
-		EndKey:   codec.EncodeInt(tablecodec.TablePrefix(), int64(0)),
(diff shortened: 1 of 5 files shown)


tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+		Id: 1,
+		Peers: []*metapb.Peer{
+			new(metapb.Peer),
+		},
+		RegionEpoch: &metapb.RegionEpoch{
+			ConfVer: 5,
+			Version: 2,
+		},
+		StartKey: codec.EncodeInt(tablecodec.TablePrefix(), int64(0)),
+		EndKey:   codec.EncodeInt(tablecodec.TablePrefix(), int64(0)),
```
coocood: The end key must be greater than start key unless both of them are empty.
Here the initial start key should be `[]byte{'t'}`, the initial end key should be `[]byte{'u'}`.

tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+	}
+}
+
+func TestKeysSplitCheck(t *testing.T) {
+	engines := newTestEngines(t)
+	defer cleanUpTestEngineData(engines)
+
+	config := newDefaultSplitCheckConfig()
+	config.regionMaxKeys = 100
+	config.regionSplitKeys = 90
```
coocood: It is `80` in tikv.

tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+		assert.Equal(t, splitRegion.RegionEpoch, expectRegion.RegionEpoch)
+		assert.Equal(t, splitRegion.SplitKeys, expectSplitKeys)
+	} else {
+		panic(fmt.Sprintf("expect split check result. but got %v", msg))
+	}
+}
+
+func putData(db *DBBundle, startIdx, endIdx uint64) {
+	val := []byte("some value")
+	arr := [1024]byte{}
```
coocood: Why not `make([]byte, 1024)`?

tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+	for startIdx < endIdx {
+		batchIdx := startIdx + 20
+		if batchIdx > endIdx {
+			batchIdx = endIdx
+		}
+		writeBatch := new(WriteBatch)
+		for i := startIdx; i < batchIdx; i++ {
+			key := DataKey(codec.EncodeBytes(nil, []byte(fmt.Sprintf("%04d", i))))
+			writeBatch.Set(key, val)
+			writeBatch.Set(key, arr[:])
```
coocood: We don't need to set the same key twice.

tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+		}
+		// Flush to generate SST files, so that properties can be utilized.
```
coocood: We don't have properties in SST, so we can just write a single batch.

tikv/raftstore/coprocessor_test.go
```diff
@@ -0,0 +1,395 @@
...
+			autoSplit: autoSplit,
+			policy:    policy,
+		},
+	})
+}
+
+func mustSplitAt(t *testing.T, rx <-chan Msg, expectRegion *metapb.Region, expectSplitKeys [][]byte) {
+	msg := <-rx
+	if msg.Type == MsgTypeRegionApproximateSize || msg.Type == MsgTypeRegionApproximateKeys {
+		assert.Equal(t, msg.RegionID, expectRegion.Id)
```
coocood: Maybe the test passed because all messsages run into this branch.

xiaobogaga:
@coocood PTAL again, thanks.

xiaobogaga:
some test are removed, because they relies on the approximation related  strategies such as `getRegionApproximateMiddle`, while these strategies are not implemented right now.

coocood:
LGTM

PR #180, opened 2019-04-17
port transport

disksing:

There are components that need to port:

- resolverRunner
- snapRunner
- raftClient


coocood:
LGTM

PR #182, opened 2019-04-18
port snap runner

disksing:

### DIFF
--- a/tikv/raftstore/snapRunner.go
+++ b/tikv/raftstore/snapRunner.go
@@ -1,13 +1,30 @@
 package raftstore
 
+import (
+	"bytes"
+	"context"
+	"io"
+	"sync/atomic"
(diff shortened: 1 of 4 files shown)


tikv/raftstore/snapRunner.go
```diff
@@ -22,9 +39,131 @@ func (r *snapRunner) run(t task) {
...
+	}
+
+	r.snapManager.Register(snapKey, SnapEntrySending)
+	defer r.snapManager.Deregister(snapKey, SnapEntrySending)
+
+	snap, err := r.snapManager.GetSnapshotForSending(snapKey)
+	if err != nil {
+		return err
+	}
+
```
coocood: Need to check `snap.Exists()`?

coocood:
LGTM

PR #185, opened 2019-04-23
 implement raftDBWriter.Write

coocood:
- Add `kvrpcpb.Context` to NewWriteBatch.
- Change Callback from function to struct type with wait group.


coocood:
@zhangjinpeng1987 @disksing PTAL

PR #186, opened 2019-04-25
implement sum agg func.

xiaobogaga:
@coocood PTAL, thanks.

### DIFF
--- a/tikv/closure_exec.go
+++ b/tikv/closure_exec.go
@@ -2,6 +2,7 @@ package tikv
 
 import (
 	"fmt"
+	"github.com/pingcap/tidb/sessionctx/stmtctx"
 	"math"
 	"sort"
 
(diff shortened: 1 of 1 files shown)


tikv/closure_exec.go
```diff
@@ -128,38 +129,50 @@ func (e *closureExecutor) initIdxScanCtx() {
...
+	default:
+		e.processFunc = &countStarProcessor{skipVal: skipVal(true), closureExecutor: e}
 	}
-	return false
+	e.finishFunc = e.countFinish
+	return e, nil
+}
+
+func (svr *Server) parseSumAggFunc(e *closureExecutor, aggExpr *tipb.Expr) (*closureExecutor, error) {
+	_, idx, err := codec.DecodeInt(aggExpr.Val)
```
coocood: The `aggExpr.Tp` is not checked, it can be a constant or scalar function.

PR #191, opened 2019-05-08
fix bugs in some test caused by code changes.

xiaobogaga:
@coocood PTAL, thanks. There are some issues in `cop_handler_test.go` and `db_writer_test.go` caused by code changes of dependent API. This pr fix this.


coocood:
LGTM

PR #192, opened 2019-05-16
Port node.rs and other minor changes

disksing:
- port node.rs
- `Bootstrap()` should return response
- update config
- `pdWorker` should create outside of raftstore
- `SnapManager` should send message using `router`


coocood:
LGTM

PR #193, opened 2019-05-24
port some local read related code

bobotu:
port code of ReadDelegate's management. Introduce `canExecuteCmdLocally`, we can use it check if a raft request is valid for current leader lease.

PTAL @coocood

### DIFF
--- a/tikv/raftstore/peer.go
+++ b/tikv/raftstore/peer.go
@@ -5,7 +5,9 @@ import (
 	"encoding/binary"
 	"fmt"
 	"math"
+	"sync/atomic"
 	"time"
+	"unsafe"
 
 	"github.com/ngaut/log"
 	"github.com/pingcap/kvproto/pkg/eraftpb"
@@ -270,6 +272,7 @@ type Peer struct {
 	PendingMergeState            *rspb.MergeState
 	leaderMissingTime            *time.Time
 	leaderLease                  *Lease
+	leaseChecker                 leaseChecker
 
 	// If a snapshot is being applied asynchronously, messages should not be sent.
 	pendingMessages         []eraftpb.Message
@@ -392,6 +395,7 @@ func (p *Peer) MaybeDestroy() *DestroyPeerJob {
 		asyncRemove = initialized
 	}
 	p.PendingRemove = true
+	p.leaseChecker.invalid.Store(true)
 
 	return &DestroyPeerJob{
 		AsyncRemove: asyncRemove,
@@ -467,7 +471,7 @@ func (p *Peer) Region() *metapb.Region {
 ///
 /// This will update the region of the peer, caller must ensure the region
 /// has been preserved in a durable device.
-func (p *Peer) SetRegion(host *CoprocessorHost, reader *readDelegate, region *metapb.Region) {
+func (p *Peer) SetRegion(host *CoprocessorHost, region *metapb.Region) {
 	if p.Region().GetRegionEpoch().GetVersion() < region.GetRegionEpoch().GetVersion() {
 		// Epoch version changed, disable read on the localreader for this region.
 		p.leaderLease.ExpireRemoteLease()
@@ -477,7 +481,7 @@ func (p *Peer) SetRegion(host *CoprocessorHost, reader *readDelegate, region *me
 	// Always update read delegate's region to avoid stale region info after a follower
 	// becoming a leader.
 	if !p.PendingRemove {
-		reader.region = region
+		atomic.StorePointer(&p.leaseChecker.region, unsafe.Pointer(region))
 		host.OnRegionChanged(p.Region(), RegionChangeEvent_Update, p.GetRole())
 	}
 }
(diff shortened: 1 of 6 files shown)


coocood:
@disksing PTAL

coocood:
@bobotu 
I think we don't need the `readDelegate`.
It's more complex than needed.

We just need a function that checks if the lease is valid. 
If not, trigger a read quorum and wait for the result.

coocood:
TiKV uses the readpool to run the read request, but unistore can just run the complete read request in the same goroutine.

disksing:
LGTM

tikv/raftstore/peer.go
```diff
@@ -474,7 +478,10 @@ func (p *Peer) SetRegion(host *CoprocessorHost, region *metapb.Region) {
 	}
 	p.Store().SetRegion(region)
 
+	// Always update read delegate's region to avoid stale region info after a follower
```
coocood: leaseChecker's region

tikv/raftstore/peer.go
```diff
@@ -834,7 +854,7 @@ func (p *Peer) PostRaftReadyAppend(ctx *PollContext, ready *raft.Ready, invokeCt
 }
 
 // Try to renew leader lease.
-func (p *Peer) MaybeRenewLeaderLease(ts time.Time) {
+func (p *Peer) MaybeRenewLeaderLease(ctx *PollContext, ts time.Time) {
```
coocood: The `ctx` is not used.

tikv/raftstore/read.go
```diff
@@ -0,0 +1,71 @@
...
+type leaseChecker struct {
+	invalid          atomic.Bool
+	peerID           atomic.Uint64
+	term             atomic.Uint64
+	appliedIndexTerm atomic.Uint64
+	leaderLease      unsafe.Pointer // *RemoteLease
+	region           unsafe.Pointer // *metapb.Region
+}
+
+func (c *leaseChecker) CheckLease(req *raft_cmdpb.RaftCmdRequest, snapTime time.Time) (bool, error) {
```
coocood: I  think it would be better to use `*kvrpcpb.Context` than `*raft_cmdpb.RaftCmdRequest`.
The caller only calls this method in read-requests, it doesn't have to compose a  `RaftCmdRequest`.

coocood: If lease expired, there should be a way to let the caller wait for the lease to update.

tikv/raftstore/read.go
```diff
@@ -23,49 +26,72 @@ type leaseChecker struct {
...
 
-type leaseCheckerInspector struct {
-	checker  *leaseChecker
-	snapTime *time.Time
-}
+	cb.wg.Wait()
 
-func (i *leaseCheckerInspector) hasAppliedToCurrentTerm() bool {
-	return i.checker.appliedIndexTerm.Load() == i.checker.term.Load()
+	return nil
```
coocood: Should check the error in `cb.resp`.

tikv/raftstore/peer.go
```diff
@@ -270,6 +272,7 @@ type Peer struct {
 	PendingMergeState            *rspb.MergeState
 	leaderMissingTime            *time.Time
 	leaderLease                  *Lease
+	leaseChecker                 leaderChecker
```
coocood: s/leaseChecker/leaderChecker

tikv/raftstore/read.go
```diff
@@ -0,0 +1,100 @@
...
+	"github.com/uber-go/atomic"
+)
+
+type LeaderChecker interface {
+	IsLeader(ctx *kvrpcpb.Context, router RaftstoreRouter) error
+}
+
+type leaderChecker struct {
+	invalid          atomic.Bool
+	peerID           atomic.Uint64
```
coocood: peerID never changes, we read it directly.

tikv/raftstore/read.go
```diff
@@ -0,0 +1,100 @@
...
+
+	if cb.resp.Header.Error != nil {
+		return &RaftError{e: cb.resp.Header.Error}
+	}
+	return nil
+}
+
+func (c *leaderChecker) isExpired(ctx *kvrpcpb.Context, snapTime *time.Time) (bool, error) {
+	if c.invalid.Load() {
+		return false, nil
```
coocood: Can we return an error when it is invalid?

tikv/raftstore/read.go
```diff
@@ -0,0 +1,100 @@
...
+	}
+
+	if ctx.Term == 0 || term <= ctx.Term+1 {
+		return false, &ErrStaleCommand{}
+	}
+
+	if ctx.RegionEpoch == nil {
+		return false, errors.New("missing epoch")
+	}
+	if ctx.RegionEpoch != region.RegionEpoch {
```
coocood: Should compare values instead of pointers.

bobotu:
PTAL @coocood

tikv/raftstore/read.go
```diff
@@ -86,7 +86,7 @@ func (c *leaderChecker) isExpired(ctx *kvrpcpb.Context, snapTime *time.Time) (bo
 	if ctx.RegionEpoch == nil {
 		return false, errors.New("missing epoch")
 	}
-	if ctx.RegionEpoch != region.RegionEpoch {
+	if ctx.RegionEpoch.Version != region.RegionEpoch.Version {
```
coocood: ConfVersion is not compared.

coocood:
LGTM

PR #195, opened 2019-05-27
raftstore: bug fixes

disksing:


disksing:
@coocood 
When a follower peer is created, the peer list in `regionMeta` is empty, and it will not be initialized until the first time it receives a snapshot.

The leader only send snapshot when it receives a message from follower, and the follower sends a message to the leader (`SendRaftMessage`) depending on the corresponding peer found.

So the initialization process will depend on `peerCache`, and when a `RaftMessage` is received, the `FromPeer` will be filled in for later use (the peer list in the regionMeta is still empty at that time).

PR #197, opened 2019-05-28
raftstore: fix snapshot and bug fixes

disksing:


tikv/raftstore/snap_builder.go
```diff
@@ -12,17 +13,17 @@ import (
...
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/tidb/util/codec"
-	"math"
 )
 
 func newSnapBuilder(cfFiles []*CFFile, db *badger.DB, lockStore *lockstore.MemStore, region *metapb.Region) (*snapBuilder, error) {
 	b := new(snapBuilder)
 	b.cfFiles = cfFiles
-	b.endKey = rawRegionKey(region.EndKey)
+	b.endKey = DataKey(rawRegionKey(region.EndKey))
```
coocood: We don't store `DataKey` prefix in the key.

coocood:
LGTM

PR #199, opened 2019-06-03
 fix snap and add log for arena

disksing:


tikv/raftstore/snap_applier.go
```diff
@@ -238,10 +238,10 @@ func readEntryFromPlainFile(data []byte) (key, value, remain []byte, err error)
...
 		return
 	}
-	key = key[1:]
-	if err != nil {
-		return
-	}
+	// key = key[1:]
+	// if err != nil {
+	// 	return
+	// }
```
coocood: Can we remove the lines?

PR #201, opened 2019-06-06
add some peer event callback

bobotu:
With these callbacks, we can maintain something like `regionCtx` and `leaderChecker` outside raftstore.

PTAL @coocood

### DIFF
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -725,9 +725,7 @@ func (d *peerFsmDelegate) destroyPeer(mergeByTarget bool) {
 		panic(d.tag() + " meta corruption detected")
 	}
 	delete(meta.regions, regionID)
-	if d.ctx.peerEventObserver != nil {
-		d.ctx.peerEventObserver.OnPeerDestroy(d.peer.getEventContext())
-	}
+	d.ctx.peerEventObserver.OnPeerDestroy(d.peer.getEventContext())
 }
 
 func (d *peerFsmDelegate) onReadyChangePeer(cp changePeer) {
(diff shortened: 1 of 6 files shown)


tikv/raftstore/fsm_peer.go
```diff
@@ -883,6 +911,10 @@ func (d *peerFsmDelegate) onReadySplitRegion(derived *metapb.Region, regions []*
 			}
 		}
 	}
+
+	if d.ctx.peerEventObserver != nil {
```
coocood: `if notifyEvent {`?

tikv/raftstore/fsm_peer.go
```diff
@@ -815,6 +832,14 @@ func (d *peerFsmDelegate) onReadySplitRegion(derived *metapb.Region, regions []*
 	// It's not correct anymore, so set it to None to let split checker update it.
 	d.peer.ApproximateSize = nil
 	lastRegionID := lastRegion.Id
+
+	notifyEvent := d.ctx.peerEventObserver != nil
```
coocood: Can we make sure the `peerEventObserver` is always not nil?

tikv/raftstore/fsm_store.go
```diff
@@ -409,6 +410,7 @@ type raftPollerBuilder struct {
 	engines              *Engines
 	applyingSnapCount    *uint64
 	tickDriverCh         chan uint64
+	peerEventObserver    PeerEventObserver
```
coocood: how can we set the `PeerEventObserver` to the builder?

bobotu:
PTAL @coocood

tikv/raftstore/node.go
```diff
@@ -23,9 +23,10 @@ type Node struct {
 	storeWg   *sync.WaitGroup
 	system    *raftBatchSystem
 	pdClient  pd.Client
+	observer  PeerEventObserver
 }
 
-func NewNode(system *raftBatchSystem, cfg *Config, pdClient pd.Client) *Node {
+func NewNode(system *raftBatchSystem, cfg *Config, pdClient pd.Client, observer PeerEventObserver) *Node {
```
coocood: observer not used.

PR #202, opened 2019-06-10
support pessimistic transaction

coocood:
Refactored based on https://github.com/ngaut/unistore/pull/173


coocood:
It now supports raft mode, but cannot run on multiple nodes for now.
Will support multi-node in the next PR.

@bobotu PTAL

PR #203, opened 2019-06-10
refactor RegionManager

bobotu:
Introduce `RegionManager` interface which supports `GetRegionFromCtx` and `CheckLeader`.

PTAL @coocood

### DIFF
--- a/tikv/region.go
+++ b/tikv/region.go
@@ -45,13 +45,16 @@ type regionCtx struct {
 
 	refCount sync.WaitGroup
 	parent   unsafe.Pointer // Parent is used to wait for all latches being released.
+
+	leaderChecker raftstore.LeaderChecker
 }
 
-func newRegionCtx(meta *metapb.Region, parent *regionCtx) *regionCtx {
+func newRegionCtx(meta *metapb.Region, parent *regionCtx, checker raftstore.LeaderChecker) *regionCtx {
 	regCtx := &regionCtx{
-		meta:    meta,
-		latches: make(map[uint64]*sync.WaitGroup),
-		parent:  unsafe.Pointer(parent),
+		meta:          meta,
+		latches:       make(map[uint64]*sync.WaitGroup),
+		parent:        unsafe.Pointer(parent),
+		leaderChecker: checker,
 	}
 	regCtx.startKey = regCtx.rawStartKey()
 	regCtx.endKey = regCtx.rawEndKey()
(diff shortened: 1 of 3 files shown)


tikv/region.go
```diff
@@ -176,10 +177,147 @@ type RegionOptions struct {
 	RegionSize int64
 }
 
-type RegionManager struct {
-	storeMeta  metapb.Store
-	mu         sync.RWMutex
-	regions    map[uint64]*regionCtx
+type RegionManager interface {
+	GetRegionFromCtx(ctx *kvrpcpb.Context) (*regionCtx, *errorpb.Error)
+	CheckLeader(ctx *kvrpcpb.Context) *errorpb.Error
```
coocood: We can do `CheckLeader` inside `GetRegionFromCtx` and remove `CheckLeader` method from `RegionManager`.

tikv/region.go
```diff
@@ -176,10 +177,147 @@ type RegionOptions struct {
...
+}
+
+func (rm *regionManager) isEpochStale(lhs, rhs *metapb.RegionEpoch) bool {
+	return lhs.GetConfVer() != rhs.GetConfVer() || lhs.GetVersion() != rhs.GetVersion()
+}
+
+type RaftRegionManager struct {
+	regionManager
+	router   raftstore.RaftstoreRouter
+	checkers map[uint64]raftstore.LeaderChecker
```
coocood: If we put `raftstore.LeaderChecker` into `regionCtx`, then we don't need to access two maps under lock.

tikv/region.go
```diff
@@ -176,10 +177,147 @@ type RegionOptions struct {
...
+	delete(rm.regions, ctx.RegionId)
+	delete(rm.checkers, ctx.RegionId)
+	rm.mu.Unlock()
+	region.refCount.Done()
+}
+
+func (rm *RaftRegionManager) OnSplitRegion(derived *metapb.Region, regions []*metapb.Region, peers []*raftstore.PeerEventContext) {
+	rm.mu.Lock()
+	oldRegion := rm.regions[derived.Id]
+	oldRegion.waitParent()
```
coocood: Can we `waitParent` outside of the lock?

coocood:
The code would be cleaner if we move the split region logic out of base `regionManager`.

bobotu:
PTAL @bobotu

coocood:
LGTM

PR #205, opened 2019-06-13
unifying the raft and stand-along server's initialization

bobotu:
Use the new `InnerServer` interface to simplify the code in `node/main.go`. 

We can remove `RunRaftServer` later.


node/main.go
```diff
@@ -111,16 +124,40 @@ func main() {
...
 	if err != nil {
 		log.Fatal(err)
 	}
-	log.Info("RegionManager closed.")
-	err = db.Close()
-	if err != nil {
-		log.Fatal(err)
+}
+
+func newRaftServerBuilder(bundle *mvcc.DBBundle, safePoint *tikv.SafePoint) server.InnerServer {
```
coocood: The returned value is not a builder.

node/main.go
```diff
@@ -66,26 +67,38 @@ func main() {
...
+		innerServer = newStandAlongBuilder(bundle, safePoint)
+	}
+
+	pdClient, err := pd.NewClient(*pdAddr, "")
+	if err != nil {
+		log.Fatal(err)
+	}
+
+	builderCtx := server.Context{
+		PDClient: pdClient,
```
coocood: How about put pdClient as an argument directly for new server function?

node/main.go
```diff
@@ -66,26 +67,38 @@ func main() {
...
+	}
+
+	if err := innerServer.Setup(&builderCtx); err != nil {
+		log.Fatal(err)
+	}
+
+	store := tikv.NewMVCCStore(bundle, *dbPath, safePoint, innerServer.GetDBWriter())
+	tikvServer := tikv.NewServer(innerServer.GetRegionManager(), store)
```
coocood: I think make the inner server as a field of `tikvServer` is better, so we can implement `TiKVServer` methods like
`func (svr *Server) Raft(tikvpb.Tikv_RaftServer) error {` with inner server.

The interface `InnerServer` can be defined in `tikv` package, the implementation of stand alone can be defined in the `tikv` package too, and `RaftInnerServer` can be defined in `raftstore` package. So we don't need the `server` package.

bobotu:
PTAL @coocood

tikv/raftstore/server.go
```diff
@@ -13,6 +13,122 @@ import (
...
+	snapManager     *SnapManager
+	coprocessorHost *CoprocessorHost
+	raftRouter      *RaftstoreRouter
+	batchSystem     *raftBatchSystem
+	pdWorker        *worker
+	resolveWorker   *worker
+	snapWorker      *worker
+}
+
+func (sb *RaftInnerServer) Raft(stream tikvpb.Tikv_RaftServer) error {
```
coocood: `sb` -> `ris`

node/main.go
```diff
@@ -66,26 +66,34 @@ func main() {
...
 	} else {
-		dbWriter = tikv.NewDBWriter(bundle, safePoint)
+		innerServer, dbWriter, regionManager = setupStandAlongInnerServer(bundle, safePoint, pdClient)
 	}
+
 	store := tikv.NewMVCCStore(bundle, *dbPath, safePoint, dbWriter)
-	tikvServer := tikv.NewServer(rm, store)
+	tikvServer := tikv.NewServer(regionManager, store, innerServer)
+
+	if err := innerServer.Start(pdClient); err != nil {
```
coocood: Can we `Start` in `setupRaftInnerServer`?

node/main.go
```diff
@@ -66,26 +66,34 @@ func main() {
...
+	}
+
+	var (
+		innerServer   tikv.InnerServer
+		dbWriter      mvcc.DBWriter
+		regionManager tikv.RegionManager
+	)
 	if *raft {
-		dbWriter = raftstore.NewDBWriter(bundle, raftstore.NewDefaultConfig())
+		innerServer, dbWriter, regionManager = setupRaftInnerServer(bundle, safePoint, pdClient)
```
coocood: We can return `store` instead of `dbWriter` here.

coocood:
LGTM

PR #209, opened 2019-06-26
fix bugs in raftstore and server

bobotu:
PTAL @coocood

### DIFF
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -39,6 +39,8 @@ type PeerEventObserver interface {
 	OnPeerDestroy(ctx *PeerEventContext)
 	// OnSplitRegion will be invoked when region split into new regions with corresponding peers.
 	OnSplitRegion(derived *metapb.Region, regions []*metapb.Region, peers []*PeerEventContext)
+	// OnRegionEpochChange will be invoked after conf change updated region's epoch.
+	OnRegionEpochChange(ctx *PeerEventContext, epoch *metapb.RegionEpoch)
 }
 
 // If we create the peer actively, like bootstrap/split/merge region, we should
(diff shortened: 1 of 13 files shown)


tikv/server.go
```diff
@@ -235,13 +250,17 @@ func (svr *Server) KvCleanup(ctx context.Context, req *kvrpcpb.CleanupRequest) (
 	if reqCtx.regErr != nil {
 		return &kvrpcpb.CleanupResponse{RegionError: reqCtx.regErr}, nil
 	}
-	err = svr.mvccStore.Cleanup(reqCtx, req.Key, req.StartVersion)
 	resp := new(kvrpcpb.CleanupResponse)
+	err = svr.mvccStore.Cleanup(reqCtx, req.Key, req.StartVersion)
```
coocood: Why move down this line?

tikv/server.go
```diff
@@ -194,16 +192,27 @@ func (p *kvScanProcessor) Process(key, value []byte) (err error) {
...
 	errs := svr.mvccStore.Prewrite(reqCtx, req.Mutations, req.PrimaryLock, req.GetStartVersion(), req.GetLockTtl())
-	return &kvrpcpb.PrewriteResponse{
-		Errors: convertToKeyErrors(errs),
-	}, nil
+
+	resp := new(kvrpcpb.PrewriteResponse)
+	var keyErrors []*kvrpcpb.KeyError
+	for _, err := range errs {
+		if err != nil {
+			if regErr := extractRegionError(err); regErr != nil {
```
coocood: There is too much boilerplate code to set region error and key error.
We can make the function return both region error and key error/errors.
Like:
```
resp.RegionError, resp.Errors = convertPBErrors(errs)
```

```
resp.RegionError, resp.Error = convertToPBError(err)
```

tikv/raftstore/fsm_peer.go
```diff
@@ -39,6 +39,8 @@ type PeerEventObserver interface {
 	OnPeerDestroy(ctx *PeerEventContext)
 	// OnSplitRegion will be invoked when region split into new regions with corresponding peers.
 	OnSplitRegion(derived *metapb.Region, regions []*metapb.Region, peers []*PeerEventContext)
+	// OnRegionEpochChange will be invoked after conf change updated region's epoch.
+	OnRegionEpochChange(ctx *PeerEventContext, epoch *metapb.RegionEpoch)
```
coocood: Can we name it `OnRegionConfChange`?
Because `SplitRegion` also changes the region epoch but doesn't call this function.

bobotu:
PTAL @coocood

coocood:
LGTM

PR #211, opened 2019-06-28
use undo log to keep lock store snapshot consistent

bobotu:
1. Introduce the undo log to lockstore.
2. During snapshot use the same `Txn` to load `ApplyState`, `RegionState` and normal database entries.

PTAL @coocood


coocood:
LGTM

PR #212, opened 2019-06-28
fix panic in onRaftGCLogTick

bobotu:
The `execCompactLog` doesn't update `aCtx.execCtx.applyState.truncatedIndex`, so `onRaftGCLogTick` will try to compact log at compacted index 😂

PTAL @coocood


coocood:
LGTM

PR #215, opened 2019-07-05
fix panic when bootstrap a bootstrapped cluster

bobotu:
PTAL @coocood

### DIFF
--- a/tikv/raftstore/node.go
+++ b/tikv/raftstore/node.go
@@ -64,9 +64,10 @@ func (n *Node) Start(ctx context.Context, engines *Engines, trans Transport, sna
 	if err != nil {
 		return err
 	}
-	if firstRegion != nil {
+	newCluster := firstRegion != nil
+	if newCluster {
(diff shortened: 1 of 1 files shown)


tikv/raftstore/node.go
```diff
@@ -197,7 +199,7 @@ func (n *Node) prepareBootstrapCluster(ctx context.Context, engines *Engines, st
 	return PrepareBootstrap(engines, storeID, regionID, peerID)
 }
 
-func (n *Node) BootstrapCluster(ctx context.Context, engines *Engines, firstRegion *metapb.Region) error {
+func (n *Node) BootstrapCluster(ctx context.Context, engines *Engines, firstRegion *metapb.Region) (bool, error) {
```
coocood: Better name the returned bool value as `newCluster`

bobotu: Use a named return value or return a local variable?

coocood:
LGTM

PR #216, opened 2019-07-09
fix mvcc and raft apply

bobotu:
Fix: 
* The `getOld` and `checkPrewrite` ignore the deleted entry, so they may return an incorrect old version or allow conflicted write.

* Prewrite of `LockKey` will delete the current value. 😂

### DIFF
--- a/tikv/raftstore/fsm_apply.go
+++ b/tikv/raftstore/fsm_apply.go
@@ -12,6 +12,7 @@ import (
 	"github.com/ngaut/unistore/tikv/mvcc"
 	"github.com/pingcap/errors"
 	"github.com/pingcap/kvproto/pkg/eraftpb"
+	"github.com/pingcap/kvproto/pkg/kvrpcpb"
 	"github.com/pingcap/kvproto/pkg/metapb"
 	"github.com/pingcap/kvproto/pkg/raft_cmdpb"
 	rspb "github.com/pingcap/kvproto/pkg/raft_serverpb"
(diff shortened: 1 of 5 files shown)


tikv/raftstore/fsm_apply.go
```diff
@@ -1117,8 +1118,10 @@ func (d *applyDelegate) execCommit(aCtx *applyContext, op commitOp) {
 	val := d.getLock(aCtx, rawKey)
 	y.Assert(len(val) > 0)
 	lock := mvcc.DecodeLock(val)
-	userMeta := mvcc.NewDBUserMeta(lock.StartTS, commitTS)
-	aCtx.wb.SetWithUserMeta(rawKey, lock.Value, userMeta)
+	if lock.Op != uint8(kvrpcpb.Op_Lock) {
```
coocood: Need to handle it in local mode too?

bobotu: Local already handle this scenario

bobotu:
IMO, we'd better use a tombstone instead delete the entry directly. Supposing this scenario:
1. `Txn1` start a scan `(a, b)` transaction at `ts1`.
2. `Txn2` deleted an entry in range `(a, b)` at commit time `ts2`.
3. `ts1` < `ts2`.
4. A network partition or any terrible network delay happened, so `Txn1` arrive after `Txn2` committed.
4. Ideally `Txn1` should read the old value of deleted value, but current scan implementation will simply skip it.

coocood:
LGTM

PR #222, opened 2019-07-31
support recover from kill -9 in raft mode

coocood:
- periodically dump lockstore entries with current raftDB vlog offset.
- recover from lockstore dump, and restore new entries from raft log.
- fix coprocessor table split checker.


coocood:
kill -9 while running sysbench read-write, after restarted, verified consistency by `admin check table`

coocood:
@bobotu PTAL

PR #224, opened 2019-08-09
replace RWMutex and map with sync.Map

coocood:
Benchmark shows it scale better on multi-core 
https://go-review.googlesource.com/c/sync/+/33912


coocood:
@bobotu PTAL

PR #226, opened 2019-08-12
support worker balance

coocood:


tikv/raftstore/peer_state.go
```diff
@@ -189,10 +204,14 @@ func (rw *raftWorker) runRaft() {
 			delegate := &peerFsmDelegate{peerFsm: peerState.peer, ctx: rw.raftCtx}
 			delegate.handleMsgs([]Msg{msg})
 		}
-		for _, peerState := range peerStateMap {
+		var movePeer uint64
+		for id, peerState := range peerStateMap {
+			movePeer = id
```
bobotu: Is it better to record the number processed message at here, so balancer can move the hottest peer to another worker?

coocood: We don't need to move the hottest peer.
A peer has some traffic is sufficient to move.

PR #227, opened 2019-08-15
replace undo log with redo raft log

coocood:
The new way to generate a consistent snapshot:

Save `redoIdx` on the peer's `applyFsm` at the beginning of the apply batch, if there is a Snapshot request, pass it all the way to `newRegionSnapshot`.

In `newRegionSnapshot`:

1. Get the region start key and end key.
2. Iterate the lockStore to generate a new lockStore for the snapshot.
3. Begin a transaction for the snapshot.
4. Get the region info again to make sure the start key and end key didn't change.
5. restore the raft log between `redoIdx` and `appliedIdx` to snapshot lockstore.

### DIFF
--- a/tikv/raftstore/fsm_apply.go
+++ b/tikv/raftstore/fsm_apply.go
@@ -247,12 +247,13 @@ func newGenSnapTask(regionID uint64, notifier chan *eraftpb.Snapshot) *GenSnapTa
 	}
 }
 
-func (t *GenSnapTask) generateAndScheduleSnapshot(regionSched chan<- task) {
+func (t *GenSnapTask) generateAndScheduleSnapshot(regionSched chan<- task, redoIdx uint64) {
 	regionSched <- task{
 		tp: taskTypeRegionGen,
 		data: &regionTask{
 			regionId: t.regionID,
 			notifier: t.snapNotifier,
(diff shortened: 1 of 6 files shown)


coocood:
@bobotu PTAL

tikv/raftstore/fsm_apply.go
```diff
@@ -552,6 +553,8 @@ type applyDelegate struct {
 	/// The term of the raft log at applied index.
 	appliedIndexTerm uint64
 
+	redoIndex uint64 // the log index that starts redo.
```
ngaut: Can not understand the comment.

tikv/raftstore/peer_state.go
```diff
@@ -16,6 +16,9 @@ type peerState struct {
 	handle unsafe.Pointer
 	peer   *peerFsm
 	apply  *applyFsm
+
+	// redoIdx is the raft log index starts redo for lockstore.
+	redoIdx uint64
```
ngaut: What's the different between redoIndex and redoIdx?

coocood: It's a redudant field, not used, I'll fix it.

PR #233, opened 2019-09-02
implement badger Guards

coocood:


coocood:
@bobotu PTAL

tikv/region.go
```diff
@@ -287,12 +287,9 @@ func (rm *RaftRegionManager) OnPeerDestroy(ctx *raftstore.PeerEventContext) {
 }
 
 func (rm *RaftRegionManager) OnSplitRegion(derived *metapb.Region, regions []*metapb.Region, peers []*raftstore.PeerEventContext) {
-	rm.mu.RLock()
```
coocood: I have got a dead-lock, many goroutines are block on `waitParent`, so I changed to `Lock`.

PR #241, opened 2019-10-11
rowcodec: fix bug in `RowToOldRow` when row is large

lysu:
when `r.large == true`,  `colIDs` will in `r.colIDs32` and `RowToOldRow` will return empty result now.

---
This change is [Reviewable](https://reviewable.io/reviews/ngaut/unistore/241)


coocood:
LGTM

PR #243, opened 2019-10-12
use ingest external table to apply snapshot

bobotu:
Update badger and use new ingest sst feature to apply snapshot


tikv/raftstore/worker.go
```diff
@@ -626,17 +637,124 @@ func newRegionRunner(engines *Engines, mgr *SnapManager, batchSize uint64, clean
 	}
 }
 
+func (r *regionRunner) tempFile() (*os.File, error) {
+	return ioutil.TempFile("", "ingest_convert_*.sst")
```
coocood: We better put the temp file under the DB dir.

PR #259, opened 2019-11-08
support multi pd server

bobotu:
PTAL @coocood


pd/client.go
```diff
@@ -77,111 +86,136 @@ type client struct {
...
-		go c.reportRegionHeartbeat(ctx, stream, errCh, wg)
-		go c.receiveRegionHeartbeat(ctx, stream, errCh, wg)
+		go c.reportRegionHeartbeat(errCh, wg)
+		go c.receiveRegionHeartbeat(errCh, wg)
 		select {
 		case err := <-errCh:
 			log.Warnf("[%s][pd] heartbeat stream get error: %s ", c.tag, err)
-			cancel()
-			time.Sleep(time.Second)
+			for {
```
coocood: The loop is too deeply nested, how about extract a method?

pd/client.go
```diff
@@ -77,111 +86,136 @@ type client struct {
...
-	cc, err := grpc.Dial(strings.TrimPrefix(c.url, "http://"), grpc.WithInsecure())
-	if err != nil {
-		return nil, errors.Trace(err)
+func (c *client) doRequest(ctx context.Context, f func(context.Context, pdpb.PDClient) error) error {
+	var err error
+	var retry int
+	for {
+		c.connMu.RLock()
+		ctx, cancel := context.WithTimeout(ctx, pdTimeout)
+		err = f(ctx, pdpb.NewPDClient(c.clientConn))
```
coocood: Can we move the f call out of the lock?

pd/client.go
```diff
@@ -61,13 +62,21 @@ var (
...
-	url        string
-	tag        string
-	clusterID  uint64
-	clientConn *grpc.ClientConn
+	url       []string
+	tag       string
+	clusterID uint64
+
+	connMu        sync.RWMutex
+	clientConn    *grpc.ClientConn
```
coocood: Can we maintain all the connections for each URL like the way https://github.com/pingcap/pd/blob/master/client/client.go do?

pd/client.go
```diff
@@ -77,111 +86,136 @@ type client struct {
...
+		c.connMu.RLock()
+		ctx, cancel := context.WithTimeout(ctx, pdTimeout)
+		err = f(ctx, pdpb.NewPDClient(c.clientConn))
+		c.connMu.RUnlock()
+		cancel()
+		if err == nil {
+			return nil
+		}
+		
+		for {
```
cfzjywxk: maybe wrap the retry logic into a `ReConnectWithRetry` function?

pd/util.go
```diff
@@ -0,0 +1,118 @@
...
+		if clusterID != 0 {
+			if cid != uint64(clusterID) {
+				return nil, nil, errors.Errorf("PD response clusterID mismatch, want %d, got %d", clusterID, cid)
+			}
+		} else {
+			clusterID = cid
+		}
+
+		if members == nil {
+			members = resp
```
cfzjywxk: could we break the loop here, if we have got the members successfully

bobotu:
PTAL @coocood @cfzjywxk

coocood:
LGTM

cfzjywxk:
LGTM better to add some comments for new functions

coocood:
@cfzjywxk we can do it later

Issue #257, opened 2019-11-04
support distributed deadlock detector


coocood:

cfzjywxk:
the distributed deadlock detect mechanism implemented in `tikv` works as follows:

After the start of tikv-server, the deadlock detecting thread running with `Detector.run` method, this thread is responsible for handling `deadlock detection` tasks as well as `cleanup` tasks.
Initially, there will be one leader tikv electing using `pd-server`,  which will handle the tasks for the whole cluster. 

The processing  data stream:
1. `acquire_pessimistic_lock` ->  `Error::KeyIsLocked` 
2. The request is added to the `waiter_manager` as a lock waiter and the wait event is triggered. 
3. Then `on_wait_for_lock` handler will handle the wait event, in which the deadlock detection will be executed, this is executed in the `LockManager.wait_for`, the detection is only used for the non-first lock. 
4.  In `handle_detect`, the detection worker will decide to process locally or to send an RPC request to the detection leader.
5. The lock wait is described as an edge like  `txnTs`(node) -> `lockTs`(node), the deadlock detection is to check if there is another edge `lockTs` -> `txnTs`, which results in loop.
6. In `do_detect`, the util will use the `wait_for_ts` as the edge start, try to find one existing edge ending with the `txn_ts`, if found there will be deadlock and report deadlock error.
7. The `cleanup` will handle wakeup event if that the lock is already acquired, and the waiting edge has also an `ttl`.

PR #263, opened 2019-11-14
support distributed dead lock detection

cfzjywxk:
Issue #257 

two related changes could be reviewed separately #269 #270 

Support distributed deadlock detection

- 1. Implement `deadlock` proto related RPC server interface
- 2. Wrap former `detector`  into new `DeadLockDetector`, add role to this new detector, and process requests based on current role(leader/follower)
- 3. Add `detectConn` and `detectClient` to do stream `send` and `recv` [this wrapping removed]
- 4. Add code path to process first region role change callback named `OnRoleChange`
- 5. Add recycling utility for detector wait map
- 6. Move the detector source code and test case from the dependent tidb package to local `tikv.detector.go` , and add some extra utilities.
- 7. Refactor lock waiter manager, remove the queue mutex

The storage engine still has some problems #267 #268,  the distributed tests are not done yet.

### DIFF
--- a/tikv/deadlock.go
+++ b/tikv/deadlock.go
@@ -18,21 +18,22 @@ const (
 	Leader
 )
 
-// DeadlockDetector is a util used for distributed deadlock detection
-type DeadlockDetector struct {
-	detector  *Detector
+type DetectorServer struct {
+	Detector *Detector
+	role     int32
+}
+
+// DetectorClient is a util used for distributed deadlock detection
+type DetectorClient struct {
 	pdClient  pd.Client
 	sendCh    chan *deadlockPb.DeadlockRequest
 	waitMgr   *lockwaiter.Manager
 	streamCli deadlockPb.Deadlock_DetectClient
-
-	// these fields used by multiple thread
-	role int32
 }
 
 // getLeaderAddr will send request to pd to find out the
 // current leader node for the first region
-func (dt *DeadlockDetector) getLeaderAddr() (string, error) {
+func (dt *DetectorClient) getLeaderAddr() (string, error) {
 	// find first region from pd, get the first region leader
 	ctx := context.Background()
 	_, leaderPeer, err := dt.pdClient.GetRegion(ctx, []byte{})
@@ -50,8 +51,8 @@ func (dt *DeadlockDetector) getLeaderAddr() (string, error) {
 }
 
 // rebuildStreamClient builds connection to the first region leader,
-// it's not thread safe and should be called only by `DeadlockDetector.Start` or `DeadlockDetector.SendReqLoop`
-func (dt *DeadlockDetector) rebuildStreamClient() error {
+// it's not thread safe and should be called only by `DetectorClient.Start` or `DetectorClient.SendReqLoop`
+func (dt *DetectorClient) rebuildStreamClient() error {
 	leaderArr, err := dt.getLeaderAddr()
 	if err != nil {
 		return err
@@ -75,27 +76,24 @@ func (dt *DeadlockDetector) rebuildStreamClient() error {
 // NewDeadlockDetector will create a new detector util, entryTTL is used for
 // recycling the lock wait edge in detector wait wap. chSize is the pending
 // detection sending task size(used on non leader node)
-func NewDeadlockDetector(waiterMgr *lockwaiter.Manager) *DeadlockDetector {
+func NewDetectorClient(waiterMgr *lockwaiter.Manager, pdClient pd.Client) *DetectorClient {
 	chSize := 10000
-	entryTTL := time.Duration(3 * time.Second)
-	urgentSize := uint64(100000)
-	exipreInterval := 3600 * time.Second
-	newDetector := &DeadlockDetector{
-		detector: NewDetector(entryTTL, urgentSize, exipreInterval),
+	newDetector := &DetectorClient{
 		sendCh:   make(chan *deadlockPb.DeadlockRequest, chSize),
 		waitMgr:  waiterMgr,
+		pdClient: pdClient,
 	}
 	return newDetector
 }
 
 // Start starts the detection `send`, `recv` and `entry recycle` loop
-func (dt *DeadlockDetector) Start() {
+func (dt *DetectorClient) Start() {
 	go dt.sendReqLoop()
 }
 
 // sendReqLoop will send detection request to leader, stream connection will be rebuilt and
 // a new recv goroutine using the same stream client will be created
-func (dt *DeadlockDetector) sendReqLoop() {
+func (dt *DetectorClient) sendReqLoop() {
 	var (
 		err        error
 		rebuildErr error
@@ -120,7 +118,7 @@ func (dt *DeadlockDetector) sendReqLoop() {
 }
 
 // recvLoop tries to recv response(current only deadlock error) from leader, break loop if errors happen
-func (dt *DeadlockDetector) recvLoop(streamCli deadlockPb.Deadlock_DetectClient) {
+func (dt *DetectorClient) recvLoop(streamCli deadlockPb.Deadlock_DetectClient) {
 	var (
 		err  error
 		resp *deadlockPb.DeadlockResponse
@@ -136,7 +134,7 @@ func (dt *DeadlockDetector) recvLoop(streamCli deadlockPb.Deadlock_DetectClient)
 	}
 }
 
-func (dt *DeadlockDetector) handleRemoteTask(requestType deadlockPb.DeadlockRequestType,
+func (dt *DetectorClient) handleRemoteTask(requestType deadlockPb.DeadlockRequestType,
 	txnTs uint64, waitForTxnTs uint64, keyHash uint64) {
 	detectReq := &deadlockPb.DeadlockRequest{}
 	detectReq.Tp = requestType
@@ -146,40 +144,21 @@ func (dt *DeadlockDetector) handleRemoteTask(requestType deadlockPb.DeadlockRequ
 	dt.sendCh <- detectReq
 }
 
-func (dt *DeadlockDetector) isLeader() bool {
-	return atomic.LoadInt32(&dt.role) == Leader
-}
-
-func (dt *DeadlockDetector) ChangeRole(newRole int32) {
-	atomic.StoreInt32(&dt.role, newRole)
-}
-
 // user interfaces
 // Cleanup processes cleaup task on local detector
-func (dt *DeadlockDetector) CleanUp(startTs uint64) {
-	if dt.isLeader() {
-		dt.detector.CleanUp(startTs)
-	} else {
-		dt.handleRemoteTask(deadlockPb.DeadlockRequestType_CleanUp, startTs, 0, 0)
-	}
+func (dt *DetectorClient) CleanUp(startTs uint64) {
+	dt.handleRemoteTask(deadlockPb.DeadlockRequestType_CleanUp, startTs, 0, 0)
 }
 
 // CleanUpWaitFor cleans up the specific wait edge in detector's wait map
-func (dt *DeadlockDetector) CleanUpWaitFor(txnTs, waitForTxn, keyHash uint64) {
-	if dt.isLeader() {
-		dt.detector.CleanUpWaitFor(txnTs, waitForTxn, keyHash)
-	} else {
-		dt.handleRemoteTask(deadlockPb.DeadlockRequestType_CleanUpWaitFor, txnTs, waitForTxn, keyHash)
-	}
+func (dt *DetectorClient) CleanUpWaitFor(txnTs, waitForTxn, keyHash uint64) {
+	dt.handleRemoteTask(deadlockPb.DeadlockRequestType_CleanUpWaitFor, txnTs, waitForTxn, keyHash)
 }
 
-// Detect will process the detection request on local detector
-func (dt *DeadlockDetector) Detect(txnTs uint64, waitForTxnTs uint64, keyHash uint64) error {
-	err := dt.detector.Detect(txnTs, waitForTxnTs, keyHash)
-	if err != nil {
-		return err
-	}
-	return nil
+// DetectRemote post the detection request to local deadlock detector or remote first region leader,
+// the caller should use `waiter.ch` to receive possible deadlock response
+func (dt *DetectorClient) Detect(txnTs uint64, waitForTxnTs uint64, keyHash uint64) {
+	dt.handleRemoteTask(deadlockPb.DeadlockRequestType_Detect, txnTs, waitForTxnTs, keyHash)
 }
 
 // convertErrToResp converts `ErrDeadlock` to `DeadlockResponse` proto type
@@ -194,16 +173,21 @@ func convertErrToResp(errDeadlock *ErrDeadlock, txnTs, waitForTxnTs, keyHash uin
 	return resp
 }
 
-// DetectRemote post the detection request to local deadlock detector or remote first region leader,
-// the caller should use `waiter.ch` to receive possible deadlock response
-func (dt *DeadlockDetector) DetectRemote(txnTs uint64, waitForTxnTs uint64, keyHash uint64) {
-	if dt.isLeader() {
-		err := dt.Detect(txnTs, waitForTxnTs, keyHash)
-		if err != nil {
-			resp := convertErrToResp(err.(*ErrDeadlock), txnTs, waitForTxnTs, keyHash)
-			dt.waitMgr.WakeUpForDeadlock(resp)
-		}
-	} else {
-		dt.handleRemoteTask(deadlockPb.DeadlockRequestType_Detect, txnTs, waitForTxnTs, keyHash)
+// NewDetectorServer creates local detector used by RPC detection handler
+func NewDetectorServer() *DetectorServer {
+	entryTTL := time.Duration(3 * time.Second)
+	urgentSize := uint64(100000)
+	exipreInterval := 3600 * time.Second
+	svr := &DetectorServer{
+		Detector: NewDetector(entryTTL, urgentSize, exipreInterval),
 	}
+	return svr
+}
+
+func (ds *DetectorServer) IsLeader() bool {
+	return atomic.LoadInt32(&ds.role) == Leader
+}
+
+func (ds *DetectorServer) ChangeRole(newRole int32) {
+	atomic.StoreInt32(&ds.role, newRole)
 }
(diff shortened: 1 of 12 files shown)


cfzjywxk:
@coocood PTAL  distributed tests not done yet, thx

tikv/region.go
```diff
@@ -313,6 +316,25 @@ func (rm *RaftRegionManager) OnRegionConfChange(ctx *raftstore.PeerEventContext,
...
+	rm.mu.RLock()
+	region, ok := rm.regions[ctx.RegionId]
+	rm.mu.RUnlock()
+	if ok {
+		if bytes.Compare(region.startKey, []byte{}) == 0 {
+			newRole := Follower
+			if newState == raft.StateLeader {
+				newRole = Leader
+			}
+			log.Infof("[ray] first region role change newRole=%v", newRole)
```
coocood: What does `ray` mean?

cfzjywxk: temporary debug log, will be removed later.

tikv/region.go
```diff
@@ -313,6 +316,25 @@ func (rm *RaftRegionManager) OnRegionConfChange(ctx *raftstore.PeerEventContext,
 	region.updateRegionEpoch(epoch)
 }
 
+func (rm *RaftRegionManager) OnRoleChange(ctx *raftstore.PeerEventContext, newState raft.StateType) {
```
coocood: We can use `regionID` as argument directly

tikv/region.go
```diff
@@ -313,6 +316,25 @@ func (rm *RaftRegionManager) OnRegionConfChange(ctx *raftstore.PeerEventContext,
...
+			newRole := Follower
+			if newState == raft.StateLeader {
+				newRole = Leader
+			}
+			log.Infof("[ray] first region role change newRole=%v", newRole)
+			rm.mvccStore.deadlockDetector.RoleChange(int32(newRole))
+		}
+	} else {
+		// should not be here
+		log.Warnf("[ray] regionid=%v not found", ctx.RegionId)
```
cfzjywxk: Is it a bug if the region map does not contain the event `regionId` ? @coocood

coocood: Yes.

tikv/deadlock.go
```diff
@@ -0,0 +1,326 @@
...
+type detectionConn struct {
+	streamMu sync.Mutex
+	stream   deadlockPb.Deadlock_DetectClient
+	ctx      context.Context
+	cancel   context.CancelFunc
+}
+
+type detectionClient struct {
+	sync.RWMutex
+	conns map[string]*detectionConn
```
coocood: We only need to maintain one connection to the leader.

cfzjywxk:
@coocood @bobotu  PTAL

tikv/server.go
```diff
@@ -562,6 +565,50 @@ func (svr *Server) UnsafeDestroyRange(context.Context, *kvrpcpb.UnsafeDestroyRan
...
+		switch req.Tp {
+		case deadlockPb.DeadlockRequestType_Detect:
+			detectErr := svr.mvccStore.deadlockDetector.Detect(req.Entry.Txn, req.Entry.WaitForTxn,
+				req.Entry.KeyHash, false, nil)
+			if detectErr != nil {
+				resp.Entry = req.Entry
+				resp.DeadlockKeyHash = detectErr.KeyHash
+			} else {
+				// TODO should we just not response?
+				resp.Entry = req.Entry
```
cfzjywxk: @coocood should we respond to caller if there is no deadlock or just let the caller waiting until timeout(3s)?

tikv/mvcc.go
```diff
@@ -291,17 +293,18 @@ func (store *MVCCStore) CheckTxnStatus(reqCtx *requestCtx,
 func (store *MVCCStore) handleCheckPessimisticErr(startTS uint64, err error, isFirstLock bool) (*lockwaiter.Waiter, error) {
 	if lock, ok := err.(*ErrLocked); ok {
 		keyHash := farm.Fingerprint64(lock.Key)
+		waiter := store.lockWaiterManager.NewWaiter(startTS, lock.StartTS, farm.Fingerprint64(lock.Key), time.Second)
```
cfzjywxk: @coocood  I moved the `NewWaiter` line in front since we need the wake-up process will send the `deadlockResponse` to the waiter's channel, any suggestions?

tikv/region.go
```diff
@@ -333,6 +336,26 @@ func (rm *RaftRegionManager) OnRegionConfChange(ctx *raftstore.PeerEventContext,
 	region.updateRegionEpoch(epoch)
 }
 
+func (rm *RaftRegionManager) OnRoleChange(regionId uint64, newState raft.StateType) {
+	rm.mu.RLock()
+	region, ok := rm.regions[regionId]
+	rm.mu.RUnlock()
+	if ok {
+		if bytes.Compare(region.startKey, []byte{}) == 0 {
```
cfzjywxk: @coocood  Is this state change processing ok?

tikv/server.go
coocood: No response if no deadlock

tikv/mvcc.go
```diff
@@ -291,17 +293,18 @@ func (store *MVCCStore) CheckTxnStatus(reqCtx *requestCtx,
...
 		if !isFirstLock { // No need to detect deadlock if it's first lock.
-			if err1 := store.deadlockDetector.Detect(startTS, lock.StartTS, keyHash); err1 != nil {
+			if err1 := store.deadlockDetector.Detect(startTS, lock.StartTS, keyHash, true, waiter); err1 != nil {
```
coocood: The Detect method should not return anything.

util/lockwaiter/lockwaiter.go
```diff
@@ -94,11 +107,12 @@ func (lw *Manager) NewWaiter(startTS, lockTS, keyHash uint64, timeout time.Durat
...
-		ch:      make(chan Result, 1),
-		startTS: startTS,
-		LockTS:  lockTS,
-		KeyHash: keyHash,
+		timeout:        timeout,
+		ch:             make(chan Result, 1),
+		startTS:        startTS,
+		LockTS:         lockTS,
+		KeyHash:        keyHash,
+		detectionResCh: make(chan *deadlock.DeadlockResponse, 1),
```
coocood: We can add a field in `Result`, and send the response to Result channel.

util/lockwaiter/lockwaiter.go
```diff
@@ -78,6 +80,17 @@ func (w *Waiter) Wait() Result {
 	}
 }
 
+// WaitDetectionRes is used for waiting response from detection leader
+// after sending detection rpc request
+func (w *Waiter) WaitDetectionRes() *deadlock.DeadlockResponse {
```
coocood: We should have only one `Wait` function.

util/lockwaiter/lockwaiter.go
```diff
@@ -145,3 +159,19 @@ func (lw *Manager) CleanUp(w *Waiter) {
 		q.removeWaiter(w)
 	}
 }
+
+// WakeUpDetection wakes up waiters waiting for deadlock detection by sending
+// rpc request to remote leader
+func (lw *Manager) WakeUpDetection(resp *deadlock.DeadlockResponse) {
```
coocood: I think `WakeUpForDeadlock` is better, we wake up the waiter because we meet a deadlock error.

tikv/deadlock.go
```diff
@@ -0,0 +1,363 @@
...
+}
+
+// user interfaces
+func (dt *DeadlockDetector) CleanUp(startTs uint64, local bool) {
+	task := &Task{taskType: CleanupTask, txnTs: startTs, isLocal: local}
+	_ = dt.handleOneTask(task)
+}
+
+func (dt *DeadlockDetector) Detect(txnTs uint64, lockTs uint64, keyHash uint64, local bool,
+	waiter *lockwaiter.Waiter) *deadlock.ErrDeadlock {
```
coocood: The method don't need the waiter argument and should return nothing.

tikv/server.go
```diff
@@ -562,6 +565,44 @@ func (svr *Server) UnsafeDestroyRange(context.Context, *kvrpcpb.UnsafeDestroyRan
...
+func (svr *Server) Detect(stream deadlockPb.Deadlock_DetectServer) error {
+	for {
+		req, err := stream.Recv()
+		if err != nil {
+			if err == io.EOF {
+				break
+			}
+			return err
+		}
+		if svr.mvccStore.deadlockDetector.isLeader() {
```
coocood: We should break the loop if it's the leader.

cfzjywxk: Break the loop if it's not the leader? @coocood

coocood: yes, my typo

tikv/deadlock.go
```diff
@@ -0,0 +1,312 @@
...
+		leaderAddr = dt.leaderStoreMeta.GetAddress()
+		err = dt.client.send(leaderAddr, req)
+		if err != nil {
+			log.Warnf("send req to addr=%v failed, err=%v", leaderAddr, err)
+			refreshErr = dt.refreshFirstRegionLeader()
+			if refreshErr != nil {
+				log.Errorf("refresh leader addr failed, err=%v", refreshErr)
+			}
+			time.Sleep(3 * time.Second)
+			continue
```
bobotu: The failed `req` will be discarded when retry?

cfzjywxk: Yes, the request will be timeout and the pessimistic lock error will be reported instead of possible deadlock error, tidb may retry the pessimistic lock request

tikv/deadlock.go
```diff
@@ -0,0 +1,312 @@
...
+
+// Follower will send detection rpc to Leader
+const (
+	Follower = iota
+	Leader
+)
+
+type detectConn struct {
+	stream deadlockPb.Deadlock_DetectClient
+	ctx    context.Context
```
coocood: We don't need to keep this field

tikv/deadlock.go
```diff
@@ -0,0 +1,312 @@
...
+	storeMeta *metapb.Store
+	sendCh    chan *deadlockPb.DeadlockRequest
+	waitMgr   *lockwaiter.Manager
+
+	// these fields should be protected by mutex
+	role            int32
+	leaderStoreMeta *metapb.Store
+}
+
+func NewDetectConn(addr string) (*detectConn, error) {
```
coocood: This method better not exported because it's only used by detectClient

tikv/deadlock.go
```diff
@@ -0,0 +1,256 @@
...
+	if err != nil {
+		log.Errorf("get first region failed, err: %v", err)
+		return err
+	}
+	leaderStoreMeta, err := dt.pdClient.GetStore(ctx, leaderPeer.GetStoreId())
+	if err != nil {
+		log.Errorf("get store=%d failed, err=%v", leaderPeer.GetStoreId(), err)
+		return err
+	}
+	dt.setLeaderStoreMeta(leaderStoreMeta)
```
coocood: We can just return the leader address instead of maintaining it as a state.

cfzjywxk:
@bobotu @coocood  PTAL
1. Detector code and test case ported from tidb dependent package, the slice is changed to list and some extra utils are added like entry size counter.
2.  `detectionConn` and `detectClient` removed.
3. Entry recycling logic added

tikv/detector.go
```diff
@@ -0,0 +1,155 @@
...
+}
+
+type txnKeyHashPair struct {
+	txn          uint64
+	keyHash      uint64
+	registerTime time.Time
+}
+
+func (p *txnKeyHashPair) isExpired(ttl time.Duration) bool {
+	if time.Since(p.registerTime) > ttl {
```
coocood: If we pass now as an argument, we can save a lot of `time.Now()` call.

cfzjywxk:
conflicts resolved.
1. Edge expire validation added in `doDetect`
2. OnRoleChange implementation based on newly region event handler goroutine

@coocood @bobotu  PTAL

util/lockwaiter/lockwaiter.go
```diff
@@ -145,3 +147,21 @@ func (lw *Manager) CleanUp(w *Waiter) {
...
+// WakeUpDetection wakes up waiters waiting for deadlock detection by sending
+// rpc request to remote leader
+func (lw *Manager) WakeUpForDeadlock(resp *deadlock.DeadlockResponse) {
+	lw.mu.Lock()
+	q := lw.waitingQueues[resp.Entry.WaitForTxn]
+	lw.mu.Unlock()
+	if q != nil {
+		for _, waiter := range q.waiters {
+			// there should be no duplicated waiters
+			if waiter.LockTS == resp.Entry.WaitForTxn && waiter.KeyHash == resp.Entry.KeyHash {
```
coocood: `waiter.LockTS == resp.Entry.WaitForTxn` this is always true.

tikv/detector.go
```diff
@@ -0,0 +1,179 @@
...
+			delete(d.waitForMap, txn)
+		}
+	}
+	d.lock.Unlock()
+
+}
+
+// activeExpire removes expired entries, should be called under d.lock protection
+func (d *Detector) activeExpire(nowTime time.Time) {
+	if nowTime.Sub(d.lastActiveExpire) > d.expireInterval ||
```
coocood: &&

util/lockwaiter/lockwaiter.go
```diff
@@ -115,56 +114,69 @@ func (lw *Manager) NewWaiter(startTS, lockTS, keyHash uint64, timeout time.Durat
...
 	}
 }
 
 // CleanUp removes a waiter from waitingQueues when wait timeout.
 func (lw *Manager) CleanUp(w *Waiter) {
 	lw.mu.Lock()
 	q := lw.waitingQueues[w.LockTS]
-	lw.mu.Unlock()
 	if q != nil {
 		q.removeWaiter(w)
```
coocood: Need to check q is empty and remove from the map

cfzjywxk: Added

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+const (
+	Follower = iota
+	Leader
+)
+
+// DeadlockDetector is a util used for distributed deadlock detection
+type DeadlockDetector struct {
+	detector  *Detector
+	pdClient  pd.Client
+	storeMeta *metapb.Store
```
coocood: This field can be removed.

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+	waitMgr   *lockwaiter.Manager
+	streamCli deadlockPb.Deadlock_DetectClient
+
+	// these fields used by multiple thread
+	role int32
+}
+
+// refreshFirstRegionLeader will send request to pd to find out the
+// current leader node for the first region
+func (dt *DeadlockDetector) refreshFirstRegionLeader() (string, error) {
```
coocood: The name better change to `getLeaderAddr`

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+		log.Errorf("get store=%d failed, err=%v", leaderPeer.GetStoreId(), err)
+		return "", err
+	}
+	log.Warnf("refreshFirstRegionLeader leader_peer=%v addr=%s", leaderPeer, leaderStoreMeta.GetAddress())
+	return leaderStoreMeta.GetAddress(), nil
+}
+
+// rebuildStreamClient builds connection to the first region leader,
+// it's not thread safe and should be called only by `DeadlockDetector.Start` or `DeadlockDetector.SendReqLoop`
+func (dt *DeadlockDetector) rebuildStreamClient() (string, error) {
```
coocood: Return the address looks weird.
We can log the address here and don't return the address.

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+	leaderArr, err := dt.refreshFirstRegionLeader()
+	if err != nil {
+		return "", err
+	}
```
coocood: We can log the leaderAddr here, then when we got an error later, we know what addr it is.

cfzjywxk: Roger, the diff intention is to log the address sending quests failed in the send loop

tikv/deadlock.go
coocood: return the streamClient better fits the method name, we can start the recvLoop goroutine in the caller code

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+		if err != nil {
+			log.Warnf("recv from failed, err=%v, stop receive", err)
+			break
+		}
+		// here only detection request will get response from leader
+		dt.waitMgr.WakeUpForDeadlock(resp)
+	}
+}
+
+func (dt *DeadlockDetector) sendReqToLeader(req *deadlockPb.DeadlockRequest) {
```
coocood: Inline this method is better.

tikv/deadlock.go
```diff
@@ -0,0 +1,217 @@
...
+func (dt *DeadlockDetector) CleanUpWaitFor(txnTs, waitForTxn, keyHash uint64) {
+	if dt.isLeader() {
+		dt.detector.CleanUpWaitFor(txnTs, waitForTxn, keyHash)
+	} else {
+		dt.handleRemoteTask(deadlockPb.DeadlockRequestType_CleanUpWaitFor, txnTs, waitForTxn, keyHash)
+	}
+}
+
+// Detect will process the detection request on local detector
+func (dt *DeadlockDetector) Detect(txnTs uint64, waitForTxnTs uint64, keyHash uint64) error {
```
coocood: Remove the local detector branch would make the code much simpler.

cfzjywxk: This is used mainly for `Server.Detect`  RPC service handler

coocood: We can separate to `DetectorClient` and `DetectorServer`, DetectorClient doesn't need to contain the `Detector` type.

cfzjywxk: `DeadlockDetector` ->(split into)  `DetectorClient`  and `DetectorServer` ,  and 
`DetectorServer`  wraps the `detector` type and do real local detection(used by rpc handler),
`DetectorClient` warps connection to the leader node and is responcible for `send` and `recv`
do i understand right?

coocood: yes

cfzjywxk: Roger, refactored

cfzjywxk: @coocood   only detection client interface kept, others removed

coocood:
LGTM

PR #269, opened 2019-11-17
fix waiter cleanup

cfzjywxk:
The lock waiter manager `Cleanup`  method should use `LockTS` to get the correspond waiting queue


cfzjywxk:
@coocood @bobotu PTAL

coocood:
LGTM

PR #270, opened 2019-11-17
pdclient: return leader from GetRegion api

cfzjywxk:
Add leader peer as return value for `GetRegion`  and `GetRegionByID`


cfzjywxk:
@coocood @bobotu  PTAL

PR #272, opened 2019-11-18
add command line argument for pd address and store address

coocood:
So we can use one config file for multiple node, set store address for different nodes, set pd address dynamically.


coocood:
@bobotu PTAL

PR #275, opened 2019-11-19
[DNM]update CheckTxnStatus API

coocood:


coocood:
@cfzjywxk PTAL

coocood:
This update makes jepsen invalid, maybe caused by the non-blocking read feature.

Issue #276, opened 2019-11-19
panic on updateRegionEpoch


coocood:
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x103bc2f]

goroutine 94 [running]:
github.com/ngaut/unistore/tikv.(*regionCtx).updateRegionEpoch(...)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/region.go:80
github.com/ngaut/unistore/tikv.(*RaftRegionManager).OnRegionConfChange(0xc000090ea0, 0xc000312700, 0xc03d2ba870)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/region.go:333 +0x7f
github.com/ngaut/unistore/tikv/raftstore.(*peerFsmDelegate).onReadyChangePeer(0xc0003a1e90, 0xc0001bda90, 0xc000085800, 0xc03d326200)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/fsm_peer.go:708 +0x22a
github.com/ngaut/unistore/tikv/raftstore.(*peerFsmDelegate).onReadyResult(0xc0003a1e90, 0x0, 0xc0002074e0, 0x1, 0x1, 0x2, 0x2, 0x40c1d6, 0xc0001305a0)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/fsm_peer.go:945 +0x40b
github.com/ngaut/unistore/tikv/raftstore.(*peerFsmDelegate).onApplyResult(0xc0003a1e90, 0xc000178ee0)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/fsm_peer.go:371 +0x17a
github.com/ngaut/unistore/tikv/raftstore.(*peerFsmDelegate).handleMsgs(0xc0003a1e90, 0xc0003a1ea0, 0x1, 0x1)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/fsm_peer.go:154 +0x780
github.com/ngaut/unistore/tikv/raftstore.(*raftWorker).run(0xc00007af00, 0xc0000dcd20, 0xc0000da590)
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/peer_state.go:153 +0x9b9
[...]

Issue #271, opened 2019-11-17
Processing region creations async may lead to invalid region id lookup in `rm.regions`


cfzjywxk:
Need investigations.

`RaftRegionManager`. `OnSplitRegion` will add `newRegionCtx` into `rm.regions` map in another goroutine,  which may lead to some event processing like `OnRoleChange` failing to get region from map using `regionId` 

Maybe we need to look closely maintaining the `regions` map

```
func (rm *RaftRegionManager) OnSplitRegion(derived *metapb.Region, regions []*metapb.Region, peers []*raftstore.PeerEventContext) {
	rm.mu.RLock()
	oldRegion := rm.regions[derived.Id]
	rm.mu.RUnlock()
	// The old region is about to be replaced, update the epoch to nil to prevent accept new requests.
	oldRegion.updateRegionEpoch(nil)
	// Run in another goroutine to avoid deadlock.
	go func() {
		oldRegion.waitParent()
		rm.mu.Lock()
		for i, region := range regions {
			rm.regions[region.Id] = newRegionCtx(region, oldRegion, peers[i].LeaderChecker)
		}
		rm.mu.Unlock()
		oldRegion.refCount.Done()
	}()
}
```

coocood:
We can run a dedicated goroutine to update the region, then all updates are executed sequentially.

PR #278, opened 2019-11-19
handle region event in a dedicated goroutine

coocood:
Fixes https://github.com/ngaut/unistore/issues/276 https://github.com/ngaut/unistore/issues/271


coocood:
@cfzjywxk PTAL

cfzjywxk:
LGTM

PR #280, opened 2019-11-21
tiny refactor on raftstore

coocood:
- Only pass necessary parameter instead of the whole PollContext to Peer, so we can infer what a method gonna do.

- remove the unused SyncLog state


coocood:
@cfzjywxk @bobotu PTAL

cfzjywxk:
LGTM

PR #283, opened 2019-11-22
snapshot apply: fix region pending apply processing

cfzjywxk:
- The `handlePendingApplies` processes apply tasks, which comes in request order in `pendingApplies` array, the successful task should be removed from the pending tasks array, otherwise, redundant processing  leads to `badger` ingestion error like `keys of external tables has overlap` 
- `TestPendingApplies` test case is fixed to make it compatible with the current interfaces

TODO, maybe we need to refactor more on the `handlePendingApplies` error processing path, and some more test cases needed to be added to verify error processing is ok(like reentrancy), to avoid breaking the data consistency

### DIFF
--- a/tikv/raftstore/worker.go
+++ b/tikv/raftstore/worker.go
@@ -828,7 +828,6 @@ func (r *regionRunner) handlePendingApplies() {
 		// Should not handle too many applies than the number of files that can be ingested.
 		// Check level 0 every time because we can not make sure how does the number of level 0 files change.
 		if r.ctx.ingestMaybeStall() {
-			log.Warnf("ingestMaybeStall break current pending applies")
 			break
 		}
 		// Try to apply task, if apply failed, throw away this task and let sender retry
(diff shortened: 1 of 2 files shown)


cfzjywxk:
@coocood @bobotu  PTAL

tikv/raftstore/worker.go
```diff
@@ -824,31 +825,38 @@ func (r *regionRunner) finishApply() error {
 
 // handlePendingApplies tries to apply pending tasks if there is some.
 func (r *regionRunner) handlePendingApplies() {
-	for _, apply := range r.pendingApplies {
+	r.leftTasks = r.leftTasks[:0]
+	for idx, apply := range r.pendingApplies {
```
coocood: How about
```go
for len(r.pendingApplies) > 0 {
    apply := r.pendingApplies[0]
    r.pendingApplies = r.pendingApplies[1:]
   ...
}

```

coocood:
LGTM

PR #284, opened 2019-11-22
fix dead lock detection for stand alone mode

bobotu:
* Close grpc connection when rebuild stream, otherwise the leaked socket will finally result in `too many open files` error.
* Set role when server starts in stand-alone mode.
* Report leader peer to PD when server start.
PTAL @coocood @cfzjywxk


cfzjywxk:
LGTM

coocood:
LGTM

PR #286, opened 2019-11-25
refactor raftstore

coocood:
- extract GlobalContext for immutable singletons.
- Separate RaftContext and StoreContext.
- Make names more consistent with their behaviors.

### DIFF
--- a/tikv/raftstore/fsm_peer.go
+++ b/tikv/raftstore/fsm_peer.go
@@ -93,8 +93,16 @@ func (pf *peerFsm) regionID() uint64 {
 	return pf.peer.regionId
 }
 
+func (d *peerFsm) region() *metapb.Region {
+	return d.peer.Store().region
+}
+
 func (pf *peerFsm) getPeer() *Peer {
 	return pf.peer
+}
+
+func (d *peerFsm) storeID() uint64 {
+	return d.peer.Meta.StoreId
 }
 
 func (pf *peerFsm) peerID() uint64 {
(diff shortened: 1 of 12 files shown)


coocood:
@cfzjywxk @bobotu PTAL

tikv/raftstore/fsm_peer.go
```diff
@@ -93,10 +93,18 @@ func (pf *peerFsm) regionID() uint64 {
...
 
+func (d *peerFsm) region() *metapb.Region {
+	return d.peer.Store().region
+}
+
 func (pf *peerFsm) getPeer() *Peer {
 	return pf.peer
 }
 
+func (d *peerFsm) storeID() uint64 {
```
cfzjywxk: Make the receiver variable name all the same?

tikv/raftstore/fsm_store.go
```diff
@@ -461,78 +375,75 @@ func (bs *raftBatchSystem) start(
...
+		snapMgr:              snapMgr,
 		router:               bs.router,
-		splitCheckScheduler:  workers.splitCheckWorker.scheduler,
-		regionScheduler:      workers.regionWorker.scheduler,
-		raftLogGCScheduler:   workers.raftLogGCWorker.scheduler,
-		compactScheduler:     workers.compactWorker.scheduler,
-		pdScheduler:          workers.pdWorker.scheduler,
-		computeHashScheduler: workers.computeHashWorker.scheduler,
 		trans:                trans,
+		pdScheduler:          bs.workers.pdWorker.scheduler,
```
cfzjywxk: Should we remove the `workers` fields and save each worker in `GlobalContext`

tikv/raftstore/fsm_store.go
```diff
@@ -62,37 +62,44 @@ func (m *storeMeta) setRegion(region *metapb.Region, peer *Peer) {
...
-	tickDriverCh         chan<- uint64
 	trans                Transport
-	queuedSnaps          map[uint64]struct{}
 	pdScheduler          chan<- task
-	raftLogGCScheduler   chan<- task
-	store                *metapb.Store
 	regionScheduler      chan<- task
-	splitCheckScheduler  chan<- task
 	computeHashScheduler chan<- task
+	raftLogGCScheduler   chan<- task
```
cfzjywxk: Should we use names like `xxxTaskQueue`  and `xxxTaskHandler/Processor`  in unistore

tikv/raftstore/fsm_store.go
```diff
@@ -62,37 +62,44 @@ func (m *storeMeta) setRegion(region *metapb.Region, peer *Peer) {
 type mergeLock struct {
 }
 
-type PollContext struct {
+type GlobalContext struct {
```
cfzjywxk: Maybe we could do more split on this?  `Global` still seems not quite direct :eyes:

tikv/raftstore/fsm_store.go
coocood: workers are not used after start, we don't need them in `GlobalContext`.

tikv/raftstore/fsm_store.go
coocood: split by what?

cfzjywxk:
LGTM maybe we need to add some unit test cases for `raftStore` related code paths

Issue #277, opened 2019-11-19
assert failed in execCommit


coocood:
2019/11/19 12:30:18 Assert failed
github.com/coocood/badger/y.Assert
	/Users/zhou/gopath/pkg/mod/github.com/coocood/badger@v1.5.1-0.20191031155124-64a2ad772588/y/error.go:55
github.com/ngaut/unistore/tikv/raftstore.(*applier).execCommit
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:1075
github.com/ngaut/unistore/tikv/raftstore.(*applier).execWriteCmd
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:883
github.com/ngaut/unistore/tikv/raftstore.(*applier).execRaftCmd
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:830
github.com/ngaut/unistore/tikv/raftstore.(*applier).applyRaftCmd
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:758
github.com/ngaut/unistore/tikv/raftstore.(*applier).processRaftCmd
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:729
github.com/ngaut/unistore/tikv/raftstore.(*applier).handleRaftEntryNormal
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:647
github.com/ngaut/unistore/tikv/raftstore.(*applier).handleRaftCommittedEntries
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:600
github.com/ngaut/unistore/tikv/raftstore.(*applier).handleApply
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:1459
github.com/ngaut/unistore/tikv/raftstore.(*applier).handleTask
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/applier.go:1552
github.com/ngaut/unistore/tikv/raftstore.(*raftWorker).runApply
	/Users/zhou/gorepo/src/github.com/ngaut/unistore/tikv/raftstore/peer_state.go:263
runtime.goexit
	/usr/local/go/src/runtime/asm_amd64.s:1357

[...]

PR #288, opened 2019-11-26
fix restore lockstore

coocood:
Fixes https://github.com/ngaut/unistore/issues/277

When unistore is killed soon after started, there is not lockstore dump file, we still need to restore the lockStore from 0 offset.


cfzjywxk:
LGTM

Issue #287, opened 2019-11-26
[RPC] Add rpc idle timeout config and refactor related logs


cfzjywxk:
After bootstrap, the unistore cluster keep printing error logs like

```
raft_client.go:107: [error] raft client failed to send
transport.go:85: [error] send raft msg err. err: EOF
```

PR #289, opened 2019-11-27
[RPC] Add keepalive for stream connection

cfzjywxk:
Solving issue #287 

Reduce the server closing stream connection times and related error logs printing, make the raft client connection keepalive take effect


cfzjywxk:
@coocood @bobotu PTAL

coocood:
Are there any cost for setting `PermitWithoutStream` to true?

cfzjywxk:
> Are there any cost for setting `PermitWithoutStream` to true?

There will be another new goroutine handling keepalive for the client,  sending ping packets to server every `time` interval if there are no packets on the stream

coocood:
LGTM

PR #291, opened 2019-11-27
[txn] Wrap some common test utility like tikv

cfzjywxk:
Add some common test transaction test utilities like in tikv, make it more convenient to port tikv transaction-related tests, tikv package now coverage now 21.9%


cfzjywxk:
@coocood @bobotu PTAL

coocood:
LGTM

PR #292, opened 2019-11-28
[txn] Complete CheckTxnStatus related logic, and port tests

cfzjywxk:
1. Let the `minCommitTs` fields take effect for prewrite and commit
2. Complete the `CheckTxnStatus` API related logic, add `Action` and `RollbackIfNotExists` field, port some tests from tikv

coverage(tikv 21.9 -> 23.8)


cfzjywxk:
@coocood  @bobotu PTAL

PR #299, opened 2019-12-02
[storage] support mvcc get by key

cfzjywxk:
Support `MvccGetByKey`  command, add this rpc handler

```
create table t1(c1 int primary key, c2 int);
insert into t1 values(1,1);
update t1 set c2 = c2 + 1 where c1 = 1;
update t1 set c2 = c2 + 50 where c1 = 1;

curl http://127.0.0.1:10080/mvcc/key/test/t1/1
{
 "key": "74800000000000002B5F728000000000000001",
 "region_id": 8,
 "value": {
  "info": {
   "writes": [
    {
     "start_ts": 412951232233078785,
     "commit_ts": 412951232233078786,
     "short_value": "CAQIaA=="
    },
    {
     "start_ts": 412951231617302529,
     "commit_ts": 412951231630147585,
     "short_value": "CAQIBA=="
    },
    {
     "start_ts": 412951231053692929,
     "commit_ts": 412951231053692930,
     "short_value": "CAQIAg=="
    }
   ],
   "values": [
    {
     "start_ts": 412951232233078785,
     "value": "CAQIaA=="
    },
    {
     "start_ts": 412951231617302529,
     "value": "CAQIBA=="
    },
    {
     "start_ts": 412951231053692929,
     "value": "CAQIAg=="
    }
   ]
  }
 }
}
```

### DIFF
--- a/tikv/dbreader/db_reader.go
+++ b/tikv/dbreader/db_reader.go
@@ -81,16 +81,12 @@ func (r *DBReader) getKeyWithMeta(key []byte, isRowKey bool, startTs uint64, mvc
 			}
 		}
 		curRecord := &kvrpcpb.MvccWrite{
-			Type:     kvrpcpb.Op_Put,
-			StartTs:  dbUsrMeta.StartTS(),
-			CommitTs: dbUsrMeta.CommitTS(),
-		}
-		curVal := &kvrpcpb.MvccValue{StartTs: curRecord.StartTs, Value: val}
-		if len(curRecord.ShortValue) <= mvcc.ShortValueMaxLen {
-			curRecord.ShortValue = val
+			Type:       kvrpcpb.Op_Put,
+			StartTs:    dbUsrMeta.StartTS(),
+			CommitTs:   dbUsrMeta.CommitTS(),
+			ShortValue: val,
 		}
 		mvccInfo.Writes = append(mvccInfo.Writes, curRecord)
-		mvccInfo.Values = append(mvccInfo.Values, curVal)
 	}
 	return nil
 }
@@ -117,16 +113,12 @@ func (r *DBReader) getOldKeysWithMeta(key []byte, startTs uint64, isRowKey bool,
 			}
 		}
 		curRecord := &kvrpcpb.MvccWrite{
-			Type:     kvrpcpb.Op_Put,
-			StartTs:  oldUsrMeta.StartTS(),
-			CommitTs: commitTs,
-		}
-		valueItem := &kvrpcpb.MvccValue{StartTs: curRecord.StartTs, Value: val}
-		if len(curRecord.ShortValue) <= mvcc.ShortValueMaxLen {
-			curRecord.ShortValue = val
+			Type:       kvrpcpb.Op_Put,
+			StartTs:    oldUsrMeta.StartTS(),
+			CommitTs:   commitTs,
+			ShortValue: val,
 		}
 		mvccInfo.Writes = append(mvccInfo.Writes, curRecord)
-		mvccInfo.Values = append(mvccInfo.Values, valueItem)
 	}
 
 	return nil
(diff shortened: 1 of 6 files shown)


cfzjywxk:
@coocood @bobotu PTAL

tikv/dbreader/db_reader.go
```diff
@@ -38,6 +42,72 @@ type DBReader struct {
...
+	item, err := r.txn.Get(key)
+	if err != nil && err != badger.ErrKeyNotFound {
+		return nil, errors.Trace(err)
+	}
+	if err == badger.ErrKeyNotFound {
+		return nil, nil
+	}
+	dbUsrMeta := mvcc.DBUserMeta(item.UserMeta())
+	if dbUsrMeta.CommitTS() <= startTs {
+		val, err := item.Value()
```
cfzjywxk: @coocood  Is it safe to use `item.Value` directly instead of `item.ValueCopy`

tikv/mvcc.go
```diff
@@ -905,6 +905,51 @@ func (store *MVCCStore) StartDeadlockDetection(ctx context.Context, pdClient pd.
...
 
+func (store *MVCCStore) MvccGetByKey(reqCtx *requestCtx, key []byte) (*kvrpcpb.MvccInfo, error) {
+	mvccInfo := &kvrpcpb.MvccInfo{}
+	lock := store.getLock(reqCtx, key)
+	if lock != nil {
+		mvccInfo.Lock = &kvrpcpb.MvccLock{
+			Type:       kvrpcpb.Op(lock.Op),
+			StartTs:    lock.StartTS,
+			Primary:    lock.Primary,
+			ShortValue: lock.Value,
```
coocood: We should avoid returning long value.

tikv/mvcc.go
```diff
@@ -905,6 +905,51 @@ func (store *MVCCStore) StartDeadlockDetection(ctx context.Context, pdClient pd.
...
+		res, err := reader.GetWithMeta(key, startTs)
+		if err != nil {
+			return nil, err
+		}
+		if res == nil {
+			break
+		}
+		val := res.Value
+		if isRowKey {
+			val, err = rowcodec.RowToOldRow(res.Value, nil)
```
coocood: We don't need to convert it to old row format.

coocood: tidb-server has the schema of the key.

tikv/dbreader/db_reader.go
bobotu: No. The returned slice may be memory-mapped, if compaction deletes the backend table, related resources will be released after transaction discarded. Which means you will get a segment fault when access value.

tikv/dbreader/db_reader.go
```diff
@@ -38,6 +42,72 @@ type DBReader struct {
...
+			StartTs:  dbUsrMeta.StartTS(),
+			CommitTs: dbUsrMeta.CommitTS(),
+			Value:    val,
+		}
+		return res, nil
+	}
+	return r.GetOldWithMeta(key, startTs)
+}
+
+func (r *DBReader) GetOldWithMeta(key []byte, startTs uint64) (*RecordWithMeta, error) {
```
coocood: This method better not export.

coocood:
We also need to look up for the rollback store for rollback entries.

tikv/dbreader/db_reader.go
coocood: Or we scan all the old versions for the key instead of lookup many times.

tikv/dbreader/db_reader.go
cfzjywxk: So we better change the `item.Value` usages in this file into `item.ValueCopy` in this file to make it safe ?  @coocood @bobotu

coocood: Yes

cfzjywxk:
> We also need to look up for the rollback store for rollback entries.

Rollback store search added

tikv/dbreader/db_reader.go
cfzjywxk: Changed to look for the current version, and then scan all historical versions

cfzjywxk:
@coocood PTAL

coocood:
LGTM

Issue #264, opened 2019-11-15
When the primary key of a transaction is Lock type, we can lost the transaction status.


coocood:
When we prewrite a LockType key, we simply remove it, the transaction status is lost.

It's OK for optimistic transactions because we never set a Lock type key as primary key.
But in pessimistic transactions, it is possible that a Lock type key is the primary key.

One solution is we change the primary key of a pessimistic transaction if it's type is Lock during the prewrite phase.

PR #301, opened 2019-12-05
fix commit Op_Lock primary key lost transaction status

coocood:
Fixes https://github.com/ngaut/unistore/issues/264

By store Op_Lock directly to old key, we can ensure the latest keys are all valid and preserve transaction status

### DIFF
--- a/tikv/dbreader/db_reader.go
+++ b/tikv/dbreader/db_reader.go
@@ -147,11 +147,11 @@ func (r *DBReader) Get(key []byte, startTS uint64) ([]byte, error) {
 	if mvcc.DBUserMeta(item.UserMeta()).CommitTS() <= startTS {
 		return item.Value()
 	}
-	item, err = r.getOldItem(key, startTS)
-	if err != nil {
-		return nil, err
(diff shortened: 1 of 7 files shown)


coocood:
@bobotu @cfzjywxk PTAL

tikv/dbreader/db_reader.go
```diff
@@ -147,27 +147,9 @@ func (r *DBReader) Get(key []byte, startTS uint64) ([]byte, error) {
...
-		// But when the latest version is DELETE and it is GCed first,
-		// we may end up here, so we should ignore the obsolete version.
-		return nil, nil
-	}
-	if nextCommitTs <= startTS {
-		// There should be a newer entry for this key, so ignore this entry.
-		return nil, nil
+	item, err = r.getOldItem(key, startTS)
+	if err != nil {
+		return nil, err
```
bobotu: If `err == badger.ErrKeyNotFound`, return `nil, nil`? Otherwise query a nonexistent key will get a `kvrpcpb.KeyError`.

Issue #300, opened 2019-12-02
support config options needed by jepsen


coocood:
https://github.com/pingcap/jepsen/blob/testing_ci/tidb/resources/tikv.conf

Those options make the test more easy to fail.

`region-max-keys` and `region-split-keys` require to implement a keySplitChecker

PR #304, opened 2019-12-10
[RaftStore] Add keys split checker

cfzjywxk:
#300 
Add keys split checker and related config fields

### DIFF
--- a/tikv/raftstore/worker.go
+++ b/tikv/raftstore/worker.go
@@ -305,15 +305,9 @@ func exceedEndKey(current, endKey []byte) bool {
 	return bytes.Compare(current, endKey) >= 0
 }
 
-/// SplitCheck gets the split keys by scanning the range.
-func (r *splitCheckHandler) splitCheck(startKey, endKey []byte, reader *dbreader.DBReader) [][]byte {
-	ite := reader.GetIter()
-	splitKeys := r.tryTableSplit(startKey, endKey, ite)
-	if len(splitKeys) > 0 {
-		return splitKeys
-	}
+// doCheck checks kvs using every checker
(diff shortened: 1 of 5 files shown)


cfzjywxk:
@coocood @bobotu  PTAL

tikv/raftstore/worker.go
```diff
@@ -295,25 +305,38 @@ func exceedEndKey(current, endKey []byte) bool {
...
+/// SplitCheck gets the split keys by scanning the range.
 func (r *splitCheckHandler) splitCheck(startKey, endKey []byte, reader *dbreader.DBReader) [][]byte {
 	ite := reader.GetIter()
 	splitKeys := r.tryTableSplit(startKey, endKey, ite)
 	if len(splitKeys) > 0 {
 		return splitKeys
 	}
-	checker := newSizeSplitChecker(r.config.regionMaxSize, r.config.regionSplitSize, r.config.batchSplitLimit)
+	r.newCheckers()
+	abort := false
```
coocood: The name `abort` is not appropriate.

tikv/raftstore/worker.go
```diff
@@ -295,25 +305,38 @@ func exceedEndKey(current, endKey []byte) bool {
...
 	for ite.Seek(startKey); ite.Valid(); ite.Next() {
 		item := ite.Item()
 		key := item.Key()
 		if exceedEndKey(key, endKey) {
 			break
 		}
-		if checker.onKv(key, uint64(item.ValueSize())) {
+		for _, checker := range r.checkers {
```
coocood: If we extract the for loop into a function, we don't need to define an `abort` variable before the outer loop.

coocood:
LGTM

PR #305, opened 2019-12-11
[coprocessor] Fix cop test

cfzjywxk:
The `kvproto` and `tipb`  changed and leads to unit test fail compliling, fix this so that the tikv package unit test will run


cfzjywxk:
@coocood  PTAL

coocood:
LGTM

PR #308, opened 2019-12-17
[config] Add config template and make it compatible with tikv conf

cfzjywxk:
Issue #300   
Add `raftstore` related configs and make them compatible with tikv,  these will be used by jepsen to make it fail fast


cfzjywxk:
@coocood @bobotu PTAL

config/config.go
```diff
@@ -35,14 +48,23 @@ type Engine struct {
...
+		RegionSize: 64 * MB,
+		LogLevel:   "info",
+		MaxProcs:   0,
+		Raft:       true,
+	},
+	RaftStore: RaftStore{
+		RaftWorkers:              2,
+		PdHeartbeatTickInterval:  20,
+		RaftStoreMaxLeaderLease:  9,
+		RaftBaseTickInterval:     1,
```
coocood: This default value is too small.

etc/config-template.toml
```diff
@@ -0,0 +1,71 @@
+## Unistore config template
```
coocood: I think it's better to put it into the `config` package.

config/config.go
```diff
@@ -35,14 +48,23 @@ type Engine struct {
...
+		StoreAddr:  "127.0.0.1:9191",
+		StatusAddr: "127.0.0.1:9291",
+		RegionSize: 64 * MB,
+		LogLevel:   "info",
+		MaxProcs:   0,
+		Raft:       true,
+	},
+	RaftStore: RaftStore{
+		RaftWorkers:              2,
+		PdHeartbeatTickInterval:  20,   // 20s
```
coocood: We better use the consistent unit, milliseconds.
Or make it clear in the name.

coocood:
LGTM

PR #309, opened 2019-12-17
Change test store to raft writer, fix found problems

cfzjywxk:
Change store to raft writer in `mvcc test`, fix two found problems

1. The applier command ops should follow input orders
eg:
```
(1) pessimistic lock one key using forUpdateTs1
(2) pessimistic the same lock using forUpdateTs2
```
Here the override (2) operation will add one `rollback` and one `put`,   but the applier did the `prewrite(the put)` first and then the `rollback(the rollback)`, this lock will be deleted unintended

2. The `minCommitTs` is not encoded into `put operation`, so after decoding, the value is incorrect

The current `make test` all passed, maybe we need to port more cases to solid the basic path

### DIFF
--- a/tikv/raftstore/restore.go
+++ b/tikv/raftstore/restore.go
@@ -70,15 +70,19 @@ func restoreAppliedEntry(entry *eraftpb.Entry, txn *badger.Txn, lockStore, rollb
 		return nil
 	}
 	writeCmdOps := createWriteCmdOps(raftCmdRequest.Requests)
-	for _, prewrite := range writeCmdOps.prewrites {
-		restorePrewrite(prewrite, txn, lockStore)
-	}
-	for _, commit := range writeCmdOps.commits {
-		restoreCommit(commit, lockStore)
-	}
-	if rollbackStore != nil {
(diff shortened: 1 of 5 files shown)


cfzjywxk:
@coocood @bobotu PTAL

tikv/raftstore/restore.go
```diff
@@ -70,15 +70,19 @@ func restoreAppliedEntry(entry *eraftpb.Entry, txn *badger.Txn, lockStore, rollb
...
-	for _, prewrite := range writeCmdOps.prewrites {
-		restorePrewrite(prewrite, txn, lockStore)
-	}
-	for _, commit := range writeCmdOps.commits {
-		restoreCommit(commit, lockStore)
-	}
-	if rollbackStore != nil {
-		for _, rollback := range writeCmdOps.rollbacks {
-			restoreRollback(rollback, rollbackStore)
+	for _, op := range writeCmdOps {
```
coocood: Why change this?

coocood:
Good job!

LGTM

PR #310, opened 2019-12-19
refactor test util, port txn tests from tikv

cfzjywxk:
~~1. Add item length check for `dbReader.get` (item.Value could be non nil zero length slice)~~
2. Port some basic txn tests from tikv(current  tikv package coverage 22.1 %)
3. Refactor some test functions and make them clean

### DIFF
--- a/tikv/cop_handler_test.go
+++ b/tikv/cop_handler_test.go
@@ -269,7 +269,7 @@ func TestPointGet(t *testing.T) {
 	// three rows data, just like the test data of table_scan.rs.
 	// then init the store with the generated data.
 	data := prepareTestTableData(t, keyNumber, TableId)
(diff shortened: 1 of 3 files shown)


cfzjywxk:
@coocood @bobotu PTAL

bobotu:
We can make badger treat `nil` and `zero-length slice` as the different values if needed. I think it will need a new flag in entry's underlaying metadata? @cfzjywxk

cfzjywxk:
> We can make badger treat `nil` and `zero-length slice` as the different values if needed. I think it need a new flag in entry's underlaying metadata? @cfzjywxk
@bobotu 

@coocood Advised using `len(item.Value()) == 0` to judge if it's empty but not using `val == nil`

cfzjywxk:
@coocood PTAL

coocood:
LGTM

PR #312, opened 2019-12-25
update badger for block cache and compression

coocood:


coocood:
@bobotu PTAL

tikv/raftstore/worker.go
```diff
@@ -811,7 +812,7 @@ func (r *regionTaskHandler) resetBuilder() error {
 		return err
 	}
 	if r.builder == nil {
-		r.builder = r.ctx.engiens.kv.DB.NewExternalTableBuilder(r.builderFile, r.ctx.mgr.limiter)
+		r.builder = r.ctx.engiens.kv.DB.NewExternalTableBuilder(r.builderFile, options.None, r.ctx.mgr.limiter)
```
bobotu: Use `config.GetGlobalConf().Engine.IngestCompression`

unistore-server/main.go
```diff
@@ -248,14 +248,17 @@ func createDB(subPath string, safePoint *tikv.SafePoint, conf *config.Engine) *b
...
-	opts.TableLoadingMode = options.MemoryMap
 	opts.ValueLogFileSize = conf.VlogFileSize
-	opts.ValueLogLoadingMode = options.FileIO
 	opts.MaxTableSize = conf.MaxTableSize
 	opts.NumMemtables = conf.NumMemTables
 	opts.NumLevelZeroTables = conf.NumL0Tables
 	opts.NumLevelZeroTablesStall = conf.NumL0TablesStall
 	opts.SyncWrites = conf.SyncWrite
+	for i := range opts.TableBuilderOptions.CompressionPerLevel {
+		opts.TableBuilderOptions.CompressionPerLevel[i] = config.ParseCompression(conf.Compression[i])
```
bobotu: Set `CompressionPerLevel` with `make([]options.CompressionType, len(conf.Compression[i])`? Otherwise we changed the value in `badger.DefaultOptions`.

unistore-server/main.go
```diff
@@ -64,6 +63,7 @@ func main() {
 	log.SetFlags(log.Ldate | log.Ltime | log.Lmicroseconds | log.Lshortfile)
 	safePoint := &tikv.SafePoint{}
 	log.Infof("conf %v", conf)
+	config.SetGlobalConf(conf)
```
bobotu: Should check the length of `compression` match the setting of `max level`?

unistore-server/main.go
coocood: No, `badger.DefaultOptions` is a struct value, its copied to `opts`.

bobotu: `CompressionPerLevel` is `[]options.CompressionType`, opts only copied the fat pointer of the slice but shared the same underlying memory.

coocood:
@bobotu PTAL

PR #313, opened 2019-12-25
fix closure executor bug

coocood:
When dag executor is TableScan->Selection->Limit, the selection is not evaluated.


coocood:
@bobotu PTAL

coocood:
@cfzjywxk PTAL

PR #317, opened 2020-01-03
support tidb's new rowformat input

lysu:
relate to https://github.com/pingcap/tidb/pull/12634

support TiDB write new format data to unionstore and always return new format data

this pr also makes unistore and tidb use same rowcode lib

---
This change is [Reviewable](https://reviewable.io/reviews/ngaut/unistore/317)


cfzjywxk:
@coocood PTAL

tikv/cop_handler.go
```diff
@@ -2,11 +2,11 @@ package tikv
 
 import (
 	"bytes"
+	"github.com/pingcap/tidb/util/codec"
```
cfzjywxk: import order

coocood:
LGTM

PR #318, opened 2020-01-06
set default split keys

coocood:
If the default value of split keys is not set, batch split would fail.


coocood:
@cfzjywxk PTAL

cfzjywxk:
LGTM

PR #319, opened 2020-01-06
lockstore supports Put operation

coocood:
So we don’t need to call Delete and Insert when update an entry.


coocood:
@bobotu PTAL

PR #322, opened 2020-01-08
support config check, add pessimistic lock wait timeout config

cfzjywxk:
1. Add lock wait timeout config, and change the default max value to `1000ms` as the latest `tikv` refactor, and tested with `tidb` repo session tests `pessimistic suite`

2. Support config check command


cfzjywxk:
@coocood @bobotu PTAL

coocood:
LGTM

PR #323, opened 2020-01-08
lockstore support PutWithHint and DeleleWithHint

coocood:
The sequential batch insert performance increased by about 50%

```
358600
376400
369000
287200
379800
377200
358200
309200
367400
377000
315600
370000
384600
374000
288400
382400
397800
365200
305800
382000
```

### DIFF
--- a/lockstore/lockstore.go
+++ b/lockstore/lockstore.go
@@ -367,26 +367,28 @@ func (ls *MemStore) calculateRecomputeHeight(key []byte, hint *Hint, listHeight 
 		recomputeHeight = listHeight
 	} else {
 		for recomputeHeight < listHeight {
-			prevNext := hint.prev[recomputeHeight].getNextNode(arena, recomputeHeight)
-			if prevNext != hint.next[recomputeHeight] {
+			prevNode := hint.prev[recomputeHeight]
+			nextNode := hint.next[recomputeHeight]
+			prevNext := prevNode.getNextNode(arena, recomputeHeight)
+			if prevNext != nextNode {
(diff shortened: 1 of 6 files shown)


coocood:
```
BenchmarkPutWithHint-12    	12651729	        95.2 ns/op
```

```
BenchmarkPut-12    	 1697431	       705 ns/op
```

lockstore/lockstore.go
```diff
@@ -326,31 +356,92 @@ func (ls *MemStore) randomHeight() int {
...
+				for bad == hint.prev[recomputeHeight] {
+					recomputeHeight++
+				}
+			} else if hint.next[recomputeHeight] != nil && bytes.Compare(key, hint.next[recomputeHeight].getKey(arena)) > 0 {
+				// Key is after splice.
+				bad := hint.next[recomputeHeight]
+				for bad == hint.next[recomputeHeight] {
+					recomputeHeight++
+				}
+			} else {
```
cfzjywxk: A bit complex with many `if` conditions in a single statement,  maybe it's easy to miss some conditions which should not go in this `else` path
how about split some condition judges
```
for recomputeHeight < listHeight {
    if  hint not valid {
        recomputeHeight++
    } else {
        if edge condition should stop {
            break
        }
        if Key is before splice {
            change recomputeHeight
        } else if  key is after splice {
            change recomputeHeight
        } else {
            found valid recomputeHeight for this key, break
        }
    }
}
```

bobotu:
LGTM

cfzjywxk:
LGTM

PR #329, opened 2020-02-07
server: return lock type when error locked

cfzjywxk:
The lock error should return lock type,  otherwise `tidb` lock resolver will get into dead loop keep reporting `txn not found`

`tps` in sysbench tests sometimes  will get into 0 and last for 20s(default lock TTL), because of the missing lock type like `pessimistic lock`


cfzjywxk:
@coocood @bobotu  PTAL

coocood:
LGTM

Issue #328, opened 2020-02-06
Pessimistic lock on already prewrite locked key not report error


cfzjywxk:
sysbench oltp_read_write test tests

txn `414441650695176250` commit secondary keys failed, and report error like

```
14:10:31.708 +08:00] [ERROR] [2pc.go:932] ["2PC failed commit key after primary key committed"] [error="[kv:8022]Error: KV error safe to retry retryable: replaced by another transaction  [try again later]"] [txnStartTS=414441650695176250] [commitTS=414441650695176310]
```

above this tidb log another txn  `414441650695176272` blocked by `414441650695176250` once pessimistic retry on this key, and this retry seems success and leads to `414441650695176250` commit error(lock has been replaced by another transaction)
```
14:10:31.706 +08:00] [INFO] [adapter.go:612] ["pessimistic write conflict, retry statement"] [conn=4] [txn=414441650695176272] [forUpdateTS=414441650695176272] [err="[kv:9007]Write conflict, txnStartTS=414441650695176272, conflictStartTS=414441650695176250, conflictCommitTS=414441650695176310, key={tableID=48, handle=1} primary=[]byte(nil) [try again later]"]
```

unistore related log

error log  `414441650695176272 ` commit failed on this key
```
14:10:31.686217 mvcc.go:663: ^[[0;31m[error] key=[116 128 0 0 0 0 0 0        48 95 114 128 0 0 0 0 0 0 1] startTS=414441650695176250 not equal lock.StartTS=       414441650695176272 commitTS=414441650695176310 lock={{414441650695176272 4144416       50695176310 0 20014 5 false 19} [116 128 0 0 0 0 0 0 54 95 114 128 0 0 0 0 0 2 1       56] [] [] []}^[[0m
```

`414441650695176272` has build pessimistic lock(which should report error)
```
14:10:31.685648 mvcc.go:250: ^[[0;37m[info] [for debug] startTS=41444       1650695176272 build pessimistic lock key=[116 128 0 0 0 0 0 0 48 95 114 128 0 0        0 0 0 0 1] lock=&{{414441650695176272 414441650695176310 0 20014 5 false 19} [11       6 128 0 0 0 0 0 0 54 95 114 128 0 0 0 0 0 2 156] [] [] []}^[[0m
```

`441650695176250 ` build prewrite lock before `414441650695176272`
```
[...]

PR #331, opened 2020-02-09
txn: check key commited from store if lock not found

cfzjywxk:
Issue #328 

Do not let `tidb` report partial commit error log during benchmark tests

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -638,16 +638,32 @@ func (store *MVCCStore) Commit(req *requestCtx, keys [][]byte, startTS, commitTS
 	var buf []byte
 	var tmpDiff int
 	var isPessimisticTxn bool
+	var lockErr error
+	var checkErr error
+	var lock mvcc.MvccLock
 	for _, key := range keys {
+		lockErr = nil
 		buf = store.lockStore.Get(key, buf)
 		if len(buf) == 0 {
 			// We never commit partial keys in Commit request, so if one lock is not found,
 			// the others keys must not be found too.
-			return store.handleLockNotFound(req, key, startTS, commitTS)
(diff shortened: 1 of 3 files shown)


cfzjywxk:
@coocood  PTAL

tikv/mvcc.go
```diff
@@ -638,16 +638,32 @@ func (store *MVCCStore) Commit(req *requestCtx, keys [][]byte, startTS, commitTS
 	var buf []byte
 	var tmpDiff int
 	var isPessimisticTxn bool
+	var lockErr error
+	var checkErr error
+	var lock mvcc.MvccLock
 	for _, key := range keys {
+		lockErr = nil
```
coocood: How about declaring variables here?

tikv/mvcc.go
```diff
@@ -639,15 +639,30 @@ func (store *MVCCStore) Commit(req *requestCtx, keys [][]byte, startTS, commitTS
...
 		}
-		lock := mvcc.DecodeLock(buf)
-		if lock.StartTS != startTS {
-			return ErrReplaced
+		if lockErr != nil {
+			// Maybe the secondary keys committed by other concurrent transactions using lock resolver,
+			// check commit info from store
+			checkErr = store.handleLockNotFound(req, key, startTS, commitTS)
+			if checkErr == nil {
+				continue
```
coocood: We can return here because the transaction is already committed.

cfzjywxk: The commit batch keys may be partially resolved by other transactions, commit batch contains secondary keys "a, b, c",  "b" is  resolved and committed by other transactions, "a, c" may still be waiting for commiting operations

coocood:
LGTM

PR #337, opened 2020-02-17
raftstore: simplify code by using single thread raftWorker.

coocood:
The multi-thread raft worker is a solution to the problem that the single thread raft worker is the bottleneck.

But things are different in Unistore:

- The raft log is in custom format, the serialization work is minimum.
- The work of writing data to DB is handled by dedicated threads maintained by the badger storage engine.
- The callback function is not executed by the apply worker.

The profile of TPCC workload shows that the CPU usage of `raftWorker` and `applyWorker` is very small. it will not be the bottleneck. So we don't need to deal with the complex logic of multi-thread raftstore.


coocood:
@cfzjywxk @bobotu PTAL

PR #350, opened 2020-03-05
fix raft store blocking on tickDriver due to channel is full

coocood:
raftWorker sends message to tickDriver and tickDriver sends message to raftWorker.
When raftWorker channel is full, the server blocking forever.


coocood:
@bobotu PTAL

PR #351, opened 2020-03-06
fix BatchGet return early when meet lock

coocood:
BatchGet should keep getting all the keys even if there are locked keys.


coocood:
@bobotu PTAL

Issue #332, opened 2020-02-10
stats: analyze table on unistore timeout


cfzjywxk:
analyze more than 10 million rows table on `unistore` using  `analyze table xxx`, the tidb rpc batch client keeps reporting
```
[WARN] [client_batch.go:594] ["wait response is cancelled"] [to=172.16.5.73:29391] [cause="context deadline exceeded"]
```
and the analyze query seems got stuck

unistore analyze processing stack
```
goroutine 96986590 [semacquire]:
sync.runtime_SemacquireMutex(0xc0000c8244, 0x0, 0x1)
	/usr/local/go/src/runtime/sema.go:71 +0x47
sync.(*Mutex).lockSlow(0xc0000c8240)
	/usr/local/go/src/sync/mutex.go:138 +0xfc
sync.(*Mutex).Lock(...)
	/usr/local/go/src/sync/mutex.go:81
math/rand.(*lockedSource).Int63(0xc0000c8240, 0x1a00)
	/usr/local/go/src/math/rand/rand.go:388 +0xe5
math/rand.(*Rand).Int63(...)
	/usr/local/go/src/math/rand/rand.go:85
math/rand.(*Rand).Int63n(0xc0000f4150, 0xb52f5, 0x2)
	/usr/local/go/src/math/rand/rand.go:117 +0x6a
math/rand.Int63n(...)
	/usr/local/go/src/math/rand/rand.go:327
github.com/pingcap/tidb/statistics.(*SampleCollector).collect(0xc010d04060, 0xc028df5680, 0x6, 0x0, 0xc02ff1e5e8, 0x2, 0x1a18, 0x0, 0x0, 0x0, ...)
	/home/tidb/go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20200103053909-324a4686f08e/statistics/sample.go:168 +0x25e
github.com/pingcap/tidb/statistics.SampleBuilder.CollectColumnStats(0xc028df5680, 0x16e6fc0, 0xc007b1aa00, 0xa, 0x0, 0x100, 0x3e8, 0x2710, 0x80000000005, 0x13, ...)
	/home/tidb/go/pkg/mod/github.com/pingcap/tidb@v1.1.0-beta.0.20200103053909-324a4686f08e/statistics/sample.go:244 +0x3fd
[...]

PR #352, opened 2020-03-09
stats: change tidb dependency using new rand function when build stats

cfzjywxk:
Solving issue #332  using new rand [function ](https://github.com/pingcap/tidb/pull/15213)in tidb,  analyze processing get much faster and will not get stock


cfzjywxk:
@coocood @bobotu PTAL

coocood:
LGTM

PR #355, opened 2020-03-12
txn: fix some cases and add check exists logic

cfzjywxk:
1. fix some case logic
2. cherry-pick op `check not exist` processing logic


cfzjywxk:
@coocood PTAL

coocood:
LGTM

PR #356, opened 2020-03-23
replace badger version with transaction TS.

coocood:
- eliminate the extra write cost caused by move old version data to another key.
- Simplify the older version data management logic.

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -1123,8 +1123,14 @@ type GCCompactionFilter struct {
 
 // Filter implements the badger.CompactionFilter interface.
 // Since we use txn ts as badger version, we do not need to filter anything.
+// It is called for the first valid version before safe point, older versions are discarded automatically.
 func (f *GCCompactionFilter) Filter(key, value, userMeta []byte) badger.Decision {
-	return badger.DecisionDrop
+	if len(value) == 0 {
+		// Delete or Op_Lock, convert it to tombstone so it can discard it if it is the most bottom level.
+		return badger.DecisionMarkTombstone
+	}
+	// Older version are discarded automatically, we need to keep the first valid version.
+	return badger.DecisionKeep
 }
 
 var (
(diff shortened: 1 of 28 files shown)


coocood:
@bobotu PTAL

tikv/raftstore/engine.go
```diff
@@ -165,53 +166,65 @@ func (wb *WriteBatch) Len() int {
...
 	wb.entries = append(wb.entries, &badger.Entry{
 		Key:      key,
 		Value:    val,
 		UserMeta: useMeta,
 	})
-	wb.size += len(key) + len(val) + len(useMeta)
+	wb.size += key.Len() + len(val) + len(useMeta)
+}
+
+func (wb *WriteBatch) SetOpLock(key y.Key, useMeta []byte) {
```
bobotu, line 209: s/useMeta/userMeta/

tikv/raftstore/engine.go
```diff
@@ -165,53 +166,65 @@ func (wb *WriteBatch) Len() int {
...
+func (wb *WriteBatch) Rollback(key y.Key) {
+	rollbackKey := mvcc.EncodeExtraTxnStatusKey(key.UserKey, key.Version)
+	wb.entries = append(wb.entries, &badger.Entry{
+		Key:      y.KeyWithTs(rollbackKey, key.Version),
+		UserMeta: mvcc.NewDBUserMeta(key.Version, 0),
 	})
 }
 
-func (wb *WriteBatch) SetWithUserMeta(key, val, useMeta []byte) {
+func (wb *WriteBatch) SetWithUserMeta(key y.Key, val, useMeta []byte) {
```
bobotu, line 200: s/useMeta/userMeta/

tikv/mvcc.go
```diff
@@ -274,6 +280,30 @@ func (store *MVCCStore) PessimisticLock(reqCtx *requestCtx, req *kvrpcpb.Pessimi
 	return nil, err
 }
 
+// extraTxnStatus can be rollback or Op_Lock that only contains transaction status info, no values.
+type extraTxnStatus struct {
+	commitTS   uint64
+	isRollback bool
```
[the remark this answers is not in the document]
coocood: No, if the status is not found, isRollback is false and commitTS is 0.

Issue #358, opened 2020-04-02
implement GreanGC methods


coocood:
To make unistore work with the latest TiDB.

PR #359, opened 2020-04-02
Fix GC and implement Green GC methods

coocood:
Fixes https://github.com/ngaut/unistore/issues/358


tikv/mvcc.go
```diff
@@ -1147,6 +1177,25 @@ func (store *MVCCStore) BatchGet(reqCtx *requestCtx, keys [][]byte, version uint
...
+		} else if lastSafePoint < safePoint {
+			store.UpdateSafePoint(safePoint)
+			lastSafePoint = safePoint
+		}
+		select {
+		case <-store.closeCh:
+			return
+		default:
+		}
+		time.Sleep(time.Minute)
```
bobotu, line 1195: Use `time.Ticker` in select?

PR #360, opened 2020-04-03
txn: skip check for force lock

cfzjywxk:
Skip the conflict check for force pessimistic lock


cfzjywxk:
@coocood PTAL

coocood:
To fully implement force lock, we also need to lock the key after wake up.

coocood:
LGTM

PR #361, opened 2020-04-03
fix Rollback doesn’t delete the lock

coocood:
- CustomLog rollback deletes a UserKey should not trim last 8 bytes.
- ScanLock should respect the limit parameter.
- ResolveLock can take given keys to avoid Scan.
- Modify the test to use CustomLog to cover the case.


coocood:
@bobotu PTAL

PR #366, opened 2020-04-03
implement Guard for raft DB

coocood:
Reduce the compaction cost for raft DB


coocood:
Seems only works on Mac, need more investigation

PR #369, opened 2020-04-07
fix apply state read stale value

coocood:
When using managed DB, we write apply state key the same version many times, it is possible that an old value is returned because it has the same version.

The solution is that we get a TS from PD at start time, put it in DBBundle, when writing to KV DB, modify the version if its version is 1.

### DIFF
--- a/tikv/raftstore/engine.go
+++ b/tikv/raftstore/engine.go
@@ -16,6 +16,7 @@ package raftstore
 import (
 	"bytes"
 	"math"
+	"sync/atomic"
 	"time"
 
 	"github.com/coocood/badger"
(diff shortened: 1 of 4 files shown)


tikv/raftstore/engine.go
```diff
@@ -251,11 +252,15 @@ func (wb *WriteBatch) RollbackToSafePoint() {
 func (wb *WriteBatch) WriteToKV(bundle *mvcc.DBBundle) error {
 	if len(wb.entries) > 0 {
 		start := time.Now()
+		keyVersion := atomic.AddUint64(&bundle.StateTS, 1)
```
bobotu, line 255: Do `RaftTS` need the same thing?

coocood: No, RaftDB is not manged.

PR #372, opened 2020-04-08
txn: complete force lock logic

cfzjywxk:
Complete the force lock logic, some tests are in [tidb-change](https://github.com/pingcap/tidb/pull/16031)


cfzjywxk:
@coocood PTAL

coocood:
LGTM

PR #376, opened 2020-04-15
extract some function for embed usage

bobotu:
Extract some code related to creating `tikv.Server` to make it easier to embed unistore into TiDB and use it like `mocktikv` to run unit-tests.


coocood:
LGTM

PR #382, opened 2020-05-13
Integrating unistore into TiDB's unit tests

bobotu:
Add MockRegionManager to provide operations needed by testkit.
We can use it in unistore's unit test.

Fixed some bugs found by TiDB's test cases. The bugs mainly appear
in coprocessor. Add some missing functions such as new collation.

The region start/end key check will cause some tests
to fail or panic, so I have removed it.


tikv/mock_region.go
```diff
@@ -0,0 +1,661 @@
...
+
+	stores := make([]*metapb.Store, 0, len(rm.stores))
+	for _, store := range rm.stores {
+		stores = append(stores, proto.Clone(store).(*metapb.Store))
+	}
+	return stores
+}
+
+// AddStore adds a new Store to the cluster.
+func (rm *MockRegionManager) AddStore(storeID uint64, addr string) {
```
bobotu, line 502: Only used by tests, no need to persist the added stores.

tikv/mock_region.go
```diff
@@ -0,0 +1,661 @@
...
+	defer rm.mu.Unlock()
+
+	rm.stores[storeID] = &metapb.Store{
+		Id:      storeID,
+		Address: addr,
+	}
+}
+
+// RemoveStore removes a Store from the cluster.
+func (rm *MockRegionManager) RemoveStore(storeID uint64) {
```
bobotu, line 513: ditto

tikv/cop_handler.go
```diff
@@ -269,8 +286,19 @@ func toPBError(err error) *tipb.Error {
 		perr.Code = int32(sqlErr.Code)
 		perr.Msg = sqlErr.Message
 	default:
-		perr.Code = int32(1)
-		perr.Msg = err.Error()
+		e := errors.Cause(err)
```
coocood, line 289: We always call `Cause`, and remove the case L284.

tikv/closure_exec.go
```diff
@@ -455,27 +484,58 @@ func (e *closureExecutor) processSelection() (gotRow bool, err error) {
...
 
+func (e *closureExecutor) copyError(err error) error {
+	if err == nil {
+		return nil
+	}
+	var ret error
+	switch x := err.(type) {
+	case *terror.Error:
+		ret = x.ToSQLError()
+	default:
```
coocood, line 527: We can remove the case at L525.

tikv/server.go
```diff
@@ -763,9 +789,5 @@ func extractRegionError(err error) *errorpb.Error {
 }
 
 func isMvccRegion(regCtx *regionCtx) bool {
```
coocood, line 791: We can remove this function.

tikv/closure_exec.go
```diff
@@ -514,6 +569,52 @@ func (e *indexScanProcessor) Finish() error {
...
+	pkStatus := e.idxScanCtx.pkStatus
+	columns := e.idxScanCtx.colInfos
+	chk := e.scanCtx.chk
+
+	colIDs := make(map[int64]int, len(columns))
+	for i, col := range columns {
+		colIDs[col.ID] = i
+	}
+	// We don't need to decode handle here, and colIDs >= 0 always.
+	rd := rowcodec.NewByteDecoder(columns, []int64{-1}, nil, nil)
```
coocood, line 589: We better reuse the row decoder.

tikv/closure_exec.go
```diff
@@ -514,6 +569,52 @@ func (e *indexScanProcessor) Finish() error {
...
+	return e.indexScanProcessOldCollation(key, value)
+}
+
+func (e *closureExecutor) indexScanProcessNewCollation(key, value []byte) error {
+	colLen := e.idxScanCtx.columnLen
+	pkStatus := e.idxScanCtx.pkStatus
+	columns := e.idxScanCtx.colInfos
+	chk := e.scanCtx.chk
+
+	colIDs := make(map[int64]int, len(columns))
```
coocood, line 584: This can be reused.

coocood:
LGTM

Issue #383, opened 2020-05-13
try separate old versions with new versions in badger SST.


coocood:
It may improve the performance of frequently updated entries.

ngaut:
To introduce table properties should be a good idea. RocksDB has something like that.

coocood:
@ngaut 
Yes, we can store table stats in the property, but I didn't get how table properties solve the old value problem?

coocood:
Do you mean set old values into table property?
If there are many old values, the property would be too large.

ngaut:
Could you store the old values into a separate SST file?

coocood:
@ngaut 
Store in a separate SST file an option, but the number of old version entries may vary a lot.
If only a few old entries, store them in another file is not efficient.

We can also reserve some blocks for the old version, those blocks are usually never read.

coocood:
The first step is done https://github.com/coocood/badger/pull/191

But it still iterates all versions.

Because we use `y.Iterator` interface to iterate values.

In order to skip old versions, all implementations of `y.Iterator` must support skip old versions.

So the next step is to change the SkipList to store old versions under the same key.

PR #390, opened 2020-05-28
update badger to skip old version and fix test

coocood:
Fixes https://github.com/ngaut/unistore/issues/383

1 table with 10million rows of sysbench table.
After 10 minutes of update-non-index, the count time reduced from 2s to 1.1s。

### DIFF
--- a/tikv/mvcc_test.go
+++ b/tikv/mvcc_test.go
@@ -1387,16 +1387,16 @@ func (s *testMvccSuite) TestResolveCommit(c *C) {
 	// The error path
 	kvTxn := store.MvccStore.db.NewTransaction(true)
 	e := &badger.Entry{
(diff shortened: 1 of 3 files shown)


tikv/mvcc_test.go
```diff
@@ -1387,16 +1387,16 @@ func (s *testMvccSuite) TestResolveCommit(c *C) {
 	// The error path
 	kvTxn := store.MvccStore.db.NewTransaction(true)
 	e := &badger.Entry{
-		Key: y.KeyWithTs(sk, 2),
+		Key: y.KeyWithTs(sk, 3),
```
coocood, line 1390: This key was written with version 2, increase the version to 3 because badger does not support update key with the same version.

PR #391, opened 2020-05-29
fix KvScan that returns early when meet lock.

coocood:
KvScan should keep scanning even if it meets lock.


coocood:
@bobotu PTAL

PR #393, opened 2020-06-02
add multi-tenant-draft-design doc

coocood:


doc/multi-tenancy.md
```diff
@@ -0,0 +1,70 @@
...
+tenant ID prefix.
+
+### Resource Isolation
+
+There is no resource isolation, all tenants shares the CPU, Memory, Disk, Network resources.
+To ensure a tenant's SLA, we limit the tenant's service quota.
+
+### Quota
+
+quota limits the resource usage of a tenant, there are three type of quota:
```
gregwebs, line 31: How is the limit actually enforced?

doc/multi-tenancy.md
```diff
@@ -0,0 +1,70 @@
...
+
+The equation looks like:
+```
+storage_quota = w1 * key_value_count + w2 * key_value size.
+```
+
+### Security
+
+- Communication layer security can be ensured by TLS.
+- SST do not do any encryption, only do encrypt on backup.
```
yiwu-arbug, line 70: we can have encryption-at-rest for the whole DB, with encryption key managed by us, just that user provided encryption key is not supported.

zhangjinpeng1987: I don't think we can decide "not support user-provided encryption key", but the PM

zhangjinpeng1987: Support or not will impact the way we store data.

doc/multi-tenancy.md
```diff
@@ -0,0 +1,70 @@
...
+quota limits the resource usage of a tenant, there are three type of quota: 
+read_quota, write_quota, storage_quota.
+
+the read_quota and write_quota is temporal, refreshed every minute.
+storage_quota is persistent.
+
+We deduct the quota after serving the tenant's request,
+When tenant's quota deducted to zero, we stop serving the request and returns an error.
+
+Every tenant has a total quota, PD is responsible to divide the total quota to each KV node that
```
gregwebs, line 40: Quotas without scheduling means under-utilization when some tenants are not using their full quota (and other tenants could be bursting to use those resources).

PR #401, opened 2020-06-09
support sample step option in KvScanRequest

coocood:
SampleStep is used for fast analyze to collect sampled key value pairs.

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -1246,8 +1246,8 @@ type kvScanProcessor struct {
 }
 
 func (p *kvScanProcessor) Process(key, value []byte) (err error) {
-	p.scanCnt++
 	if p.sampleStep > 0 {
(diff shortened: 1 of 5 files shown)


tikv/mvcc.go
```diff
@@ -1239,11 +1239,19 @@ func isResolved(startTS uint64, resolved []uint64) bool {
...
-	pairs []*kvrpcpb.KvPair
+	buf        []byte
+	pairs      []*kvrpcpb.KvPair
+	sampleStep uint32
+	scanCnt    uint32
 }
 
 func (p *kvScanProcessor) Process(key, value []byte) (err error) {
+	p.scanCnt++
+	if p.sampleStep > 0 {
```
lzmhhh123, line 1250:
```suggestion
	if p.sampleStep > 0 {
		p.scanCnt++
```

PR #402, opened 2020-06-23
lockstore: compatible with Go 1.14's checkptr

bobotu:
Just like https://github.com/pingcap/tidb/pull/14972


bobotu:
PTAL @coocood

coocood:
LGTM

PR #404, opened 2020-06-30
tikv: add server closed check in SplitRegion and KvGC

bobotu:
These two handlers don't acquire server ref count and check `closed` flag.
This will cause panic like https://github.com/pingcap/tidb/issues/18236


bobotu:
PTAL @coocood

coocood:
LGTM

PR #407, opened 2020-07-15
txn: implement some async commit related logic

cfzjywxk:
Try to implement these async commit protocol related features:
- Update `kvproto`.
- Return `minCommitTS` in `Prewrite` response, fetching new ts from PD.
- `CheckTxnStatus` related change.
- Return `minCommitTS` in lock error information.

Tests:
- Unit tests.
- Simple demo with TiDB changes using async commit protocol.

### DIFF
--- a/tikv/mvcc/mvcc.go
+++ b/tikv/mvcc/mvcc.go
@@ -27,25 +27,37 @@ type DBUserMeta []byte
 
 const dbUserMetaLen = 16
 
-// DecodeLock decodes data to lock, the primary and value is copied.
+// DecodeLock decodes data to lock, the primary and value is copied, the secondaries are copied if async commit is enabled.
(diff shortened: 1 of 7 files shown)


cfzjywxk:
@coocood PTAL

tikv/mvcc/mvcc.go
```diff
@@ -27,46 +27,73 @@ type DBUserMeta []byte
...
-	TTL         uint32
-	Op          uint8
-	HasOldVer   bool
-	PrimaryLen  uint16
+	StartTS        uint64
+	ForUpdateTS    uint64
+	MinCommitTS    uint64
+	TTL            uint32
+	Op             uint8
+	HasOldVer      bool
```
coocood, line 57: The `HasOldVer` is not used anymore, we can use it as `AsyncCommit` flag.

PR #409, opened 2020-08-03
implement CheckSecondaryLockStatus

sticnarf:


sticnarf:
@cfzjywxk @coocood PTAL

tikv/mvcc.go
```diff
@@ -451,6 +451,55 @@ func (store *MVCCStore) CheckTxnStatus(reqCtx *requestCtx,
...
+					Key:             key,
+					LockTtl:         uint64(lock.TTL),
+					LockType:        kvrpcpb.Op(lock.Op),
+					LockForUpdateTs: lock.ForUpdateTS,
+					UseAsyncCommit:  lock.UseAsyncCommit,
+					MinCommitTs:     lock.MinCommitTS,
+				})
+			}
+		} else {
+			commitTS, err := store.checkCommitted(reqCtx.getDBReader(), key, startTS)
```
coocood, line 489: May be the key is already rolled back, we should call `checkExtraTxnStatus`.

sticnarf:
FIXME: This PR forgets to wake up rolled back locks.

tikv/mvcc.go
sticnarf: I've updated the PR and add the Op_Lock test case. PTAL again.

coocood:
LGTM

coocood:
@cfzjywxk PTAL

cfzjywxk:
LGTM

PR #412, opened 2020-08-10
update badger for better windows support

bobotu:
Update badger to include https://github.com/pingcap/badger/pull/202


bobotu:
PTAL @coocood

coocood:
LGTM

PR #414, opened 2020-08-20
update pd dependencies

HunDunDM:
### What problem does this PR solve? 

This PR is the stage 1 of https://github.com/pingcap/tidb/issues/19306 , in order to upgrade the PD dependency of tidb related repo.

**Note**: After merging this PR, it will temporarily be unable to compile.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -14,11 +14,11 @@ require (
 	github.com/pingcap/errors v0.11.5-0.20200729012136-4e113ddee29e
 	github.com/pingcap/kvproto v0.0.0-20200810113304-6157337686b1
 	github.com/pingcap/log v0.0.0-20200511115504-543df19646ad
-	github.com/pingcap/pd/v4 v4.0.0-rc.2.0.20200730093003-dc8c75cf7ca0
 	github.com/pingcap/tidb v1.1.0-beta.0.20200806060043-574540aa06ba
 	github.com/prometheus/client_golang v1.5.1
 	github.com/shirou/gopsutil v2.19.10+incompatible
 	github.com/stretchr/testify v1.5.1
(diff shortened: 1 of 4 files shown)


coocood:
```
panic: duplicate metrics collector registration attempted

goroutine 1 [running]:
github.com/prometheus/client_golang/prometheus.(*Registry).MustRegister(0xc0001bcb90, 0xc00007cbb0, 0x1, 0x1)
	/Users/zhou/gopath/pkg/mod/github.com/prometheus/client_golang@v1.5.1/prometheus/registry.go:400 +0xad
github.com/prometheus/client_golang/prometheus.MustRegister(...)
	/Users/zhou/gopath/pkg/mod/github.com/prometheus/client_golang@v1.5.1/prometheus/registry.go:177
github.com/pingcap/pd/v4/client.init.0()
	/Users/zhou/gopath/pkg/mod/github.com/pingcap/pd/v4@v4.0.0-rc.2.0.20200730093003-dc8c75cf7ca0/client/metrics.go:85 +0x91
```
maybe we need to remove the MustRegister in pd client init function?

HunDunDM:
This is because the current project still indirectly depends on `pingcap/pd/v4` because of `tidb` dependency, which will be resolved in stage 3.

coocood:
If tikv/pd/client remove the register, we can upgrade smoothly.

HunDunDM:
Relying on two versions of PD is not expected behavior. We always need the stage 3 PR to update tidb.

coocood:
My point is that we better not break the build during the upgrade process.

coocood:
Or we can create a branch of tikv pd client that doesn't register, used for repos depends on pingcap pd.

HunDunDM:
Will be merge to the `update-pd-tmp` branch

PR #417, opened 2020-08-25
txn: fix scan lock process

cfzjywxk:
The scan processor needs to consider the `limit` input parameter for the data scan and lock scan(thanks for @bobotu for reminding about this key point). 
Currently, the lock scan may return more keys than expected by the input limit parameter, if there are keys being committed in the same range.

Try to add the limit parameter into `collectRangeLock` and the `validPairs` collecting logic so there will not be more keys than expected.

It's easy to reproduce the scan_test failure on TiDB skipping commit specific secondary keys, like this:
```
skip these two keys doing commit, log in unistore
[WARN] [mvcc.go:865] ["[for debug] skip commit key"] [startTS=418999667188039680] [key=7480000000000000015f7280000000000002f5]
[WARN] [mvcc.go:865] ["[for debug] skip commit key"] [startTS=418999667188039680] [key=7480000000000000015f7280000000000002fc]

the scan_test failed
[ERROR] [scan_test.go:79] ["bytes equal check fail"] [i=754] [rowNum=768] ["obtained key"=7480000000000000015f7280000000000002f5] ["obtained val"=80373537] [expected=7480000000000000015f7280000000000002f2]

the next scan result is the skipped lock keys, they are not continuous 
[INFO] [scan_test.go:88] [key] [key=7480000000000000015f7280000000000002f5] [j=754]
[2020/08/25 20:01:24.748 +08:00] 
[INFO] [scan_test.go:88] [key] [key=7480000000000000015f7280000000000002fc] [j=755]


```

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -1306,10 +1306,9 @@ func (store *MVCCStore) BatchGet(reqCtx *requestCtx, keys [][]byte, version uint
 	return pairs
 }
 
-func (store *MVCCStore) collectRangeLock(startTS uint64, startKey, endKey []byte, resolved []uint64, limit int) []*kvrpcpb.KvPair {
+func (store *MVCCStore) collectRangeLock(startTS uint64, startKey, endKey []byte, resolved []uint64) []*kvrpcpb.KvPair {
 	var pairs []*kvrpcpb.KvPair
 	it := store.lockStore.NewIterator()
-	var cnt int
 	for it.Seek(startKey); it.Valid(); it.Next() {
 		if exceedEndKey(it.Key(), endKey) {
 			break
@@ -1321,10 +1320,6 @@ func (store *MVCCStore) collectRangeLock(startTS uint64, startKey, endKey []byte
 				Error: convertToKeyError(err),
 				Key:   safeCopy(it.Key()),
 			})
-		}
-		cnt++
-		if cnt >= limit {
-			break
 		}
 	}
 	return pairs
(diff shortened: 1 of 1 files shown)


coocood:
LGTM

coocood:
@bobotu PTAL

tikv/mvcc.go
```diff
@@ -1306,9 +1306,10 @@ func (store *MVCCStore) BatchGet(reqCtx *requestCtx, keys [][]byte, version uint
 	return pairs
 }
 
-func (store *MVCCStore) collectRangeLock(startTS uint64, startKey, endKey []byte, resolved []uint64) []*kvrpcpb.KvPair {
+func (store *MVCCStore) collectRangeLock(startTS uint64, startKey, endKey []byte, resolved []uint64, limit int) []*kvrpcpb.KvPair {
 	var pairs []*kvrpcpb.KvPair
 	it := store.lockStore.NewIterator()
+	var cnt int
 	for it.Seek(startKey); it.Valid(); it.Next() {
```
bobotu, line 1313: Should we handle ReverseScan at here? It seems like this procedure always scan locks ascending, so it probably doesn't work properly in reverse mode now.
If we add a count limit here, we must handle reverse scan explicitly, because if we simply swap the `startKey` and `endKey` when invoke this funtion, it will skip some keys which we actually need it.

cfzjywxk: Seems the `lockstore` iterator does not support reverse scan by now, maybe we could scan the whole range locks and only keep the valid pair collection limit check.

bobotu: We can use `Iterator.Prev` for ReverseScan. Because the `startKey`is the largest key bound in ReverseScan mode.

bobotu: But use `Prev` is a bit complicated, scan the whole range seems better.

cfzjywxk: Got it.

bobotu (commented):
LGTM

bobotu (approved):
LGTM

PR #419, opened 2020-09-03
hide go.mod file to avoid dependency on every history version

coocood:
Because tidb and unistore is has cyclic dependency, go mod recursively check out every history version of tidb and unistore and add them in go.sum. If one version is missing, the build would fail.

So we hide the go.mod file in repository to break it. like the way https://github.com/pingcap/tidb-tools does.


cfzjywxk:
LGTM

PR #421, opened 2020-09-18
remove the lock ts set for waiter wakeup to avoid race

cfzjywxk:
Try to fix the issue https://github.com/pingcap/tidb/issues/19868, remove the `LockTS` set for wake up.


cfzjywxk:
@coocood PTAL

util/lockwaiter/lockwaiter.go
```diff
@@ -189,7 +189,6 @@ func (lw *Manager) WakeUp(txn, commitTS uint64, keyHashes []uint64) {
 	// wake up delay waiters, this will not remove waiter from queue
 	if len(wakeUpDelayWaiters) > 0 {
 		for _, w := range wakeUpDelayWaiters {
-			w.LockTS = txn
```
coocood, line 192: Why it is added in the first place?

cfzjywxk: I don't remember the reason at that time..

PR #422, opened 2020-09-28
do not allow primary get to bypass async commit locks

sticnarf:
Port https://github.com/tikv/tikv/pull/8598
This change guarantees that a point get can read the record which has just been written using async commit.


sticnarf:
PTAL @coocood

coocood:
LGTM

PR #423, opened 2020-09-29
Fix missing lock fields in KeyError and forbid CheckTxnStatus pushing async commit locks

sticnarf:
CheckTxnStatus is changed to be aligned with TiKV.

I've tested locally that with this fix, the tests in https://github.com/pingcap/tidb/pull/20073 can pass.

### DIFF
--- a/tikv/mvcc.go
+++ b/tikv/mvcc.go
@@ -512,10 +512,11 @@ func (store *MVCCStore) normalizeWaitTime(lockWaitTime int64) time.Duration {
 }
 
 func (store *MVCCStore) handleCheckPessimisticErr(startTS uint64, err error, isFirstLock bool, lockWaitTime int64) (*lockwaiter.Waiter, error) {
-	if lock, ok := err.(*ErrLocked); ok {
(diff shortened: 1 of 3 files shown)


tikv/mvcc.go
```diff
@@ -934,7 +938,7 @@ func (store *MVCCStore) rollbackKeyReadLock(reqCtx *requestCtx, batch mvcc.Write
 		}
 		if lock.StartTS == startTS {
 			if currentTs > 0 && uint64(oracle.ExtractPhysical(lock.StartTS))+uint64(lock.TTL) >= uint64(oracle.ExtractPhysical(currentTs)) {
-				return rollbackStatusLocked, BuildLockErr(key, key, lock.StartTS, uint64(lock.TTL), lock.Op, lock.MinCommitTS)
```
sticnarf, line 937: I think previously the primary key field was set wrong.

sticnarf:
@coocood PTAL

coocood:
LGTM

PR #424, opened 2020-10-14
Add support for single region 1pc

MyonKeminta:
Signed-off-by: MyonKeminta <MyonKeminta@users.noreply.github.com>

### DIFF
--- a/go.mod1
+++ b/go.mod1
@@ -12,7 +12,7 @@ require (
 	github.com/pingcap/badger v1.5.1-0.20200908111422-2e78ee155d19
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
 	github.com/pingcap/errors v0.11.5-0.20200729012136-4e113ddee29e
-	github.com/pingcap/kvproto master
+	github.com/pingcap/kvproto v0.0.0-20201020051409-cdcb788eaebd
 	github.com/pingcap/log v0.0.0-20200511115504-543df19646ad
 	github.com/pingcap/tidb v1.1.0-beta.0.20200820085534-0d997f2b8b3c
 	github.com/prometheus/client_golang v1.5.1
--- a/go.mod1
+++ b/go.mod1
@@ -12,7 +12,7 @@ require (
 	github.com/pingcap/badger v1.5.1-0.20200908111422-2e78ee155d19
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
 	github.com/pingcap/errors v0.11.5-0.20200729012136-4e113ddee29e
-	github.com/pingcap/kvproto v0.0.0-20200927054727-1290113160f0
+	github.com/pingcap/kvproto master
 	github.com/pingcap/log v0.0.0-20200511115504-543df19646ad
 	github.com/pingcap/tidb v1.1.0-beta.0.20200820085534-0d997f2b8b3c
 	github.com/prometheus/client_golang v1.5.1
--- a/go.mod1
+++ b/go.mod1
@@ -12,7 +12,7 @@ require (
 	github.com/pingcap/badger v1.5.1-0.20200908111422-2e78ee155d19
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
 	github.com/pingcap/errors v0.11.5-0.20200729012136-4e113ddee29e
-	github.com/pingcap/kvproto v0.0.0-20200828054126-d677e6fd224a
+	github.com/pingcap/kvproto v0.0.0-20200927054727-1290113160f0
 	github.com/pingcap/log v0.0.0-20200511115504-543df19646ad
 	github.com/pingcap/tidb v1.1.0-beta.0.20200820085534-0d997f2b8b3c
 	github.com/prometheus/client_golang v1.5.1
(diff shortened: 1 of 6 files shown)


tikv/mvcc.go
```diff
@@ -682,17 +697,23 @@ func (store *MVCCStore) prewritePessimistic(reqCtx *requestCtx, mutations []*kvr
...
 		physical, logical, tsErr := store.pdClient.GetTS(context.Background())
 		if tsErr != nil {
 			return tsErr
 		}
 		minCommitTS = uint64(physical)<<18 + uint64(logical)
-		reqCtx.asyncMinCommitTS = minCommitTS
+		if req.UseAsyncCommit {
+			reqCtx.asyncMinCommitTS = minCommitTS
+		} else if !req.TryOnePc {
+			log.Fatal("unreachable")
```
coocood, line 711: This branch is obviously unreachable, we can just remove it.

tikv/mvcc.go
```diff
@@ -682,17 +697,23 @@ func (store *MVCCStore) prewritePessimistic(reqCtx *requestCtx, mutations []*kvr
...
+		}
 	}
+
+	keys := make([][]byte, 0, len(mutations))
```
coocood, line 715: Can we avoid memory allocation if not OnePC?

tikv/mvcc.go
```diff
@@ -682,17 +697,23 @@ func (store *MVCCStore) prewritePessimistic(reqCtx *requestCtx, mutations []*kvr
...
+	locks := make([]*mvcc.MvccLock, 0, len(mutations))
 	for i, m := range mutations {
 		if m.Op == kvrpcpb.Op_CheckNotExists {
 			continue
```
coocood, line 719: unfold the code below, we don't need to set the MinCOmmitTs in the loop.

MyonKeminta: It looks needed for `buildPrewriteLock`.

coocood: Yes, but we can put it out of the loop.

tikv/mvcc.go
```diff
@@ -704,11 +725,54 @@ func (store *MVCCStore) prewriteMutations(reqCtx *requestCtx, mutations []*kvrpc
...
+	}
+	if minCommitTS < req.StartVersion {
+		log.Fatal("1pc commitTS less than startTS", zap.Uint64("startTS", req.StartVersion), zap.Uint64("minCommitTS", minCommitTS))
+	}
+
+	reqCtx.onePCCommitTS = minCommitTS
+	store.updateLatestTS(minCommitTS)
+	batch := store.dbWriter.NewWriteBatch(req.StartVersion, minCommitTS, reqCtx.rpcCtx)
+	for i, key := range keys {
+		batch.Commit(key, locks[i])
```
MyonKeminta, line 767: There's one more problem. Trying to commit a key that is not locked will cause panicking. Looks like there need a special method for `WriteBatch` to implement it.

MyonKeminta: I found that there's a complete raft system which saves the commit message in raftlog, which also need to be changed to support it. @coocood Do you think it's a good idea to change them all? or is there any better idea?

coocood: We can support raft mode later.

coocood: We can add a bool in `MvccLockHdr` for onePC.
And do not delete the lock if it is onePC.

MyonKeminta: It sometimes needs to delete the pessimistic lock.

coocood: We could also check if the lock is pessimistic.

coocood:
@MyonKeminta why the diff is so large?

MyonKeminta:
Something goes wrong. Wait a minute. Sorry

tikv/mvcc/mvcc.go
```diff
@@ -59,6 +60,8 @@ type MvccLockHdr struct {
 	PrimaryLen     uint16
 	UseAsyncCommit bool
 	SecondaryNum   uint32
+	UseOnePC       bool
```
coocood, line 63: Move the two fields before `SecondaryNum` can save storage.

tikv/mvcc/mvcc.go
```diff
@@ -59,6 +60,8 @@ type MvccLockHdr struct {
...
+	IsPessimistic  bool
```
coocood, line 64: We don't need this field, we can check the Op instead.

MyonKeminta: This happens when prewriting. Can the `Op` show whether the key is locked by a pessimistic lock? As I understood it, it only distinguishes the type of the modification like `Put`, `Delete`, `Insert`, `Lock`.

coocood: No, the pessimistic lock has its own code `Op_PessimisticLock Op = 5`.

MyonKeminta:
I think maybe it's better to merge this later, since I think @sticnarf 's latest proposal may need to update kvproto again

MyonKeminta:
kvproto was updated

go.mod1
```diff
@@ -12,7 +12,7 @@ require (
 	github.com/pingcap/badger v1.5.1-0.20200908111422-2e78ee155d19
 	github.com/pingcap/check v0.0.0-20200212061837-5e12011dc712
 	github.com/pingcap/errors v0.11.5-0.20200729012136-4e113ddee29e
-	github.com/pingcap/kvproto v0.0.0-20200927054727-1290113160f0
+	github.com/pingcap/kvproto master
```
coocood, line 15: Why not use a specific version?

MyonKeminta: Oh, sorry, I ran `make test` before committing and it doesn't copy go.mod back to go.mod1

coocood:
LGTM

PR #425, opened 2020-10-22
pd,tikv: replace juju/errors by pingcap/errors

kennytm:
Unistore still contain some old code using `juju/errors` rather than only `pingcap/errors`. This PR replaced all remaining `juju/errors` by `pingcap/errors`, so that TiDB can be free of `juju/errors` after updating the go.mod entry.


kennytm:
PTAL @coocood

coocood:
LGTM

PR #427, opened 2020-11-13
migrate tiflash mock test from mocktikv to unistore

windtalker:
migrate tiflash mock test from mocktikv to unistore


coocood:
LGTM

PR #430, opened 2020-12-09
support mpp tests

windtalker:


tikv/region.go
```diff
@@ -246,8 +246,10 @@ type RegionOptions struct {
 }
 
 type RegionManager interface {
-	GetRegionFromCtx(ctx *kvrpcpb.Context) (*regionCtx, *errorpb.Error)
+	GetRegionFromCtx(ctx *kvrpcpb.Context) (*regionCtx, *errorpb.Error, string, uint64)
```
coocood, line 249: Can we use a separate method to get the store address and id?

windtalker: OK, updated.

tikv/server.go
```diff
@@ -558,7 +586,16 @@ func (svr *Server) BatchCoprocessor(req *coprocessor.BatchRequest, batchCopServe
 		if reqCtx.regErr != nil {
 			return &RegionError{err: reqCtx.regErr}
 		}
-		copResponse := cophandler.HandleCopRequest(reqCtx.getDBReader(), svr.mvccStore.lockStore, &cop)
+		var mppTaskHandler *cophandler.MPPTaskHandler = nil
```
coocood, line 589: `= nil` can be removed

tikv/server.go
```diff
@@ -515,7 +534,16 @@ func (svr *Server) Coprocessor(ctx context.Context, req *coprocessor.Request) (*
 	if reqCtx.regErr != nil {
 		return &coprocessor.Response{RegionError: reqCtx.regErr}, nil
 	}
-	return cophandler.HandleCopRequest(reqCtx.getDBReader(), svr.mvccStore.lockStore, req), nil
+	var mppTaskHandler *cophandler.MPPTaskHandler = nil
+	if mockRegionRM, ok := svr.regionManager.(*MockRegionManager); ok {
+		mppTaskHandlerMap := mockRegionRM.getMPPTaskSet(reqCtx.storeId)
```
coocood, line 539: Need to check nil?

tikv/server.go
```diff
@@ -515,7 +534,16 @@ func (svr *Server) Coprocessor(ctx context.Context, req *coprocessor.Request) (*
...
+		if th, ok := mppTaskHandlerMap[int64(req.Context.TaskId)]; ok {
+			mppTaskHandler = th
+		}
+	}
+	return cophandler.HandleCopRequestWithMPPCtx(reqCtx.getDBReader(), svr.mvccStore.lockStore, req, &cophandler.MPPCtx{
```
coocood, line 544: Has `HandleCopRequest` been removed?
Can we call the old function if there is no mpp task?

windtalker: Not, but `HandleCopRequest` just call `HandleCopRequestWithMPPCtx` by passing `nil`  as the `MPPCtx`. Here, we do not call `HandleCopRequest`  because in the implementation, `mppHandler.HandleMPPDispatch` will call `BatchCoprocessor` or `Coprocessor` to handle the dag request, so inside `BatchCoprocessor`/`Coprocessor` there is no way to know if it is just a `BatchCoprocessor`/`Coprocessor` or actually a mpp task.

coocood:
LGTM

PR #434, opened 2021-02-01
add lock gaurd for mpp task loading and storing to avoid data race

hanfei1991:
As the mpp test increases, there are tasks running on same store. We must add lock to avoid data race.

### DIFF
--- a/tikv/server.go
+++ b/tikv/server.go
@@ -534,14 +534,15 @@ func (svr *Server) Coprocessor(_ context.Context, req *coprocessor.Request) (*co
 	if reqCtx.regErr != nil {
 		return &coprocessor.Response{RegionError: reqCtx.regErr}, nil
 	}
-	var mppTaskHandler *cophandler.MPPTaskHandler
 	if mockRegionRM, ok := svr.regionManager.(*MockRegionManager); ok {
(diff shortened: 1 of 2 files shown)


tikv/server.go
```diff
@@ -591,18 +592,7 @@ func (svr *Server) BatchCoprocessor(req *coprocessor.BatchRequest, batchCopServe
...
-		var mppTaskHandler *cophandler.MPPTaskHandler
-		if mockRegionRM, ok := svr.regionManager.(*MockRegionManager); ok {
-			mppTaskHandlerMap := mockRegionRM.getMPPTaskSet(reqCtx.storeId)
-			if mppTaskHandlerMap != nil {
-				if th, ok := mppTaskHandlerMap[int64(req.Context.TaskId)]; ok {
-					mppTaskHandler = th
-				}
-			}
-		}
-		copResponse := cophandler.HandleCopRequestWithMPPCtx(reqCtx.getDBReader(), svr.mvccStore.lockStore, &cop, &cophandler.MPPCtx{
```
coocood, line 603: So the `MPPCtx` is not needed?

hanfei1991: Yeah, "Batch Coprocessor" has no bessiness of MPP, I think that's a mistake of previous design.

coocood:
LGTM

PR #436, opened 2021-02-18
tikv/server: refine mpp code without mess

hanfei1991:
we could run mpp task directly instead of converting it to cop requests.


coocood:
LGTM

PR #440, opened 2021-03-04
adjust the position of finish req ctx in mpp request

hanfei1991:
the finish of reqCtx will close the reader thoroughly. So we have to finish it after the execution ends.


coocood:
LGTM

PR #445, opened 2021-04-15
Set up actions ci and fix check errors.

hslam:
* Updated unistore config.
* Migrated tikv/raftstore to raftstore.
* Fixed check errors.
* Added check to makefile.
* Set up github actions ci.


coocood:
LGTM

PR #585, opened 2021-07-15
speed up recovery

hslam:


engine/engine.go
```diff
@@ -358,45 +359,84 @@ func formatInt(n int) string {
 }
 
 func (en *Engine) loadShards() error {
-	for _, mShard := range en.manifest.shards {
+	sche := scheduler.NewScheduler(en.opt.RecoveryConcurrency)
+	var parents = make(map[uint64]struct{})
+	parentsBT := scheduler.NewBatchTasks()
+	hasParentBT := scheduler.NewBatchTasks()
```
coocood, line 365: Can we use the same batch for hasParent shards and normal shards?

hslam: fixed

coocood:
LGTM

PR #627, opened 2021-08-17
use l0 key as the split key when the l0 tables size is larger than 30%

hslam:
* Use l0 key as the split key when the l0 tables size is larger than 30%.
* Fix bug where recovery state could not continue to split.


coocood:
LGTM

PR #664, opened 2021-09-18
fix the incorrect peer state

hslam:


tikv/raftstore/fsm_peer.go
```diff
@@ -716,7 +716,9 @@ func (d *peerMsgHandler) onReadyChangePeer(cp changePeer) {
 		}
 		delete(d.peer.followersSplitFilesDone, peerID)
 		d.peer.removePeerCache(peerID)
-		WritePeerState(d.ctx.raftWB, cp.region, rspb.PeerState_Tombstone, nil)
+		WritePeerState(d.ctx.raftWB, cp.region, rspb.PeerState_Normal, nil)
```
coocood, line 719: Maybe we can just remove this line?
The according code in TiKV doesn't write peer state here.

hslam: But in TiKV, the peer state is written on *[store/fsm/apply.rs](https://github.com/tikv/tikv/blob/762531710c8eb0a9f97b3a2d4b3bb929f413c74d/components/raftstore/src/store/fsm/apply.rs#L1908-L1915)*.
```
                region.mut_peers().push(peer.clone());
                PEER_ADMIN_CMD_COUNTER_VEC
                    .with_label_values(&["add_learner", "success"])
                    .inc();
                info!(
                    "add learner successfully";
                    "region_id" => self.region_id(),
                    "peer_id" => self.id(),
                    "peer" => ?peer,
                    "region" => ?&self.region
                );
            }
        }
        let state = if self.pending_remove {
            PeerState::Tombstone
        } else {
            PeerState::Normal
        };
        if let Err(e) = write_peer_state(ctx.kv_wb_mut(), &region, state, None) {
            panic!("{} failed to update region state: {:?}", self.tag, e);
        }
        let mut resp = AdminResponse::default();
        resp.mut_change_peer().set_region(region.clone());
```
There was similar code in unistore before.
https://github.com/ngaut/unistore/blob/cd951cdb5473dc11d7cee6890825cba324419104/tikv/raftstore/fsm_apply.go#L1226-L1236
But now, no peer state is written on applier. Instead of being written on peer msg handler. So we still need to write peer state.
https://github.com/ngaut/unistore/blob/ac57495647b8bbb0f490d5b1c97992cdb4aee277/tikv/raftstore/applier.go#L982-L987

PR #667, opened 2021-09-24
add raft proposal and process metrics

hslam:


tikv/raftstore/applier.go
```diff
@@ -525,33 +528,33 @@ func (a *applier) handleRaftEntryConfChange(aCtx *applyContext, entry *eraftpb.E
...
 	peerID := a.peer.Id
 	if isConfChange {
 		cmd := a.pendingCmds.takeConfChange()
 		if cmd == nil {
-			return nil
+			return nil, 0
 		}
 		if cmd.index == index && cmd.term == term {
-			return cmd.cb
+			return cmd.cb, time.Now().Sub(cmd.startTime).Seconds()
```
coocood, line 540: This can be done in the `Callback.Done`, So we can keep this method signature unchanged.

tikv/raftstore/peer_worker.go
```diff
@@ -250,6 +250,7 @@ func (rw *raftWorker) handleRaftReady(readyRes []*ReadyICPair) {
 			h.HandleRaftReady(&pair.Ready, pair.IC)
 		}
 		rw.handleReadyDc.collect(time.Since(begin))
+		metrics.PeerRaftProcessDuration.WithLabelValues("ready").Observe(time.Now().Sub(begin).Seconds())
```
coocood, line 253: We can reuse the `time.Since(begin)` value at line 252.

tikv/raftstore/worker.go
```diff
@@ -446,6 +446,7 @@ func (r *regionTaskHandler) handle(t task) {
 	regionTask := t.data.(*regionTask)
 	switch t.tp {
 	case taskTypeRegionApply:
+		metrics.ApplyTaskWaitTimeHistogram.Observe(time.Now().Sub(regionTask.startTime).Seconds())
```
coocood, line 449: use `time.Since(regionTask.startTime)` is simpler.

tikv/raftstore/applier.go
hslam: Fixed.

tikv/raftstore/peer_worker.go
hslam: Updated.

tikv/raftstore/worker.go
hslam: Updated.

PR #668, opened 2021-09-28
add raft command metrics

hslam:


tikv/raftstore/peer.go
```diff
@@ -251,6 +261,8 @@ type Peer struct {
 	applyProposals []proposal
 	pendingReads   *ReadIndexQueue
 
+	commandsTracker map[uint64]*commandTrace
```
coocood, line 264: `commandsTraces` would be more consistent.

tikv/raftstore/peer_worker.go
```diff
@@ -163,29 +168,33 @@ func (rw *raftWorker) receiveMsgs(closeCh <-chan struct{}) (quit bool) {
...
 		if msg.Type == MsgTypeRaftMessage {
 			raftMsgCount++
 		}
 		rw.getPeerInbox(msg.RegionID).append(msg)
 	case msg := <-rw.applyResCh:
+		receivingTime = time.Now()
 		respCount++
 		rw.getPeerInbox(msg.RegionID).append(msg)
 	case <-rw.ticker.C:
+		receivingTime = time.Now()
```
coocood, line 190: It would be simpler to get the time after the `select` block.

hslam: Fixed it.

PR #673, opened 2021-10-12
make raft engine thread-safe

coocood:
Prepare for implementing multiple raft worker.

### DIFF
--- a/raftengine/engine.go
+++ b/raftengine/engine.go
@@ -131,11 +131,15 @@ func (e *Engine) Write(wb *WriteBatch) error {
 	for _, op := range wb.raftLogOps {
 		e.writer.appendRaftLog(op)
 	}
-	e.mapMu.RLock()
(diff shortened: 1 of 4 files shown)


raftengine/engine.go
```diff
@@ -106,51 +110,71 @@ type stateOp struct {
...
 		e.worker.taskCh <- rotateTask{epochID: epochID, states: states}
 	}
 	for _, op := range wb.raftLogOps {
 		e.writer.appendRaftLog(op)
+	}
+	e.mapMu.RLock()
+	for _, op := range wb.raftLogOps {
 		getRegionRaftLogs(e.entriesMap, op.regionID).append(op)
 	}
+	e.mapMu.RUnlock()
```
hslam, line 138: In the `getRegionRaftLogs` function, if the `RegionRaftLogs` does not exist in the `entriesMap`, it will be created and stored in the map, and it may be created repeatedly with read lock.

